PR #47, opened 2019-12-04
Allow custom subnets in canaries

RedbackThomson:
Creating a new EKS cluster for every canary run requires configuring a new VPC and subnets each time. Sometimes CloudFormation fails to delete the cluster CloudFormation stacks because the VPCs don't release. This meant that canaries were failing due to issues with VPC account limits. 

This CR adds the option to specify existing subnets (connected to existing VPCs) that it will pass through to `eksctl` when creating the stacks. By doing so, we are able to manage the VPCs, subnets, routing tables, internet gateways and any other VPC resources ourselves - ensuring we do not reach account limits.

### DIFF
--- a/tests/codebuild/run_canarytest.sh
+++ b/tests/codebuild/run_canarytest.sh
@@ -5,7 +5,11 @@
 # Build environment `Docker image` has all prerequisite setup and credentials are being passed using AWS system manager
 
 CLUSTER_REGION=${CLUSTER_REGION:-us-east-1}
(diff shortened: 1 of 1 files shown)


RedbackThomson:
> LGTM, just curious what would be the impact of this on users ?

I have left the defaults to be that the functionality remains the same for anyone who previously relied on `eksctl` to create their VPC. That is, if you don't specify the subnets, it creates them for you.

PR #49, opened 2019-12-05
Updating v1.0 installer image to ECR v1.0 tag

cadedaniel:

I tested this by deploying it to my cluster. I created a TrainingJob and everything worked as expected.

@RedbackThomson I ran into troubles updating the v1.0 branch because it is the same name as the tag, can we switch to use `r1.0` for branches and `v1.0` for tags?  This is how [Tensorflow](https://github.com/tensorflow/tensorflow/branches) does it.


RedbackThomson:
Sounds great. Lets do it!

PR #50, opened 2019-12-06
Float release SEMVER up to major and minor tags

RedbackThomson:
## Description
When tagging with `vX.Y.Z`, the `vX` and `vX.Y` tags will now also be updated to the latest version. This means we are now able to offer customers with major, minor and patch specific tags.

## Testing
Ran this on our development CI/CD pipeline

### DIFF
--- a/codebuild/scripts/release_tag.sh
+++ b/codebuild/scripts/release_tag.sh
@@ -18,10 +18,10 @@ function retag_image()
   docker push $image:$new_tag
 }
 
-CODEBUILD_GIT_TAG="$(git describe --tags --exact-match 2>/dev/null)"
+GIT_TAG="$(git describe --tags --exact-match 2>/dev/null)"
 
 # Only run the release process for tagged commits
-if [ "$CODEBUILD_GIT_TAG" == "" ]; then
+if [ "$GIT_TAG" == "" ]; then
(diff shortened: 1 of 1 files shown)


codebuild/scripts/release_tag.sh
```diff
@@ -18,10 +18,10 @@ function retag_image()
   docker push $image:$new_tag
 }
 
-CODEBUILD_GIT_TAG="$(git describe --tags --exact-match 2>/dev/null)"
+GIT_TAG="$(git describe --tags --exact-match 2>/dev/null)"
```
cadedaniel: Will we have more than one tag now? e.g. v1.1 -> v1.1.6 -> commit sha?

RedbackThomson: In git we will only have `vX.Y.Z` tags, but in the containers and artifacts we will have each of the types.

cadedaniel: Ah, I see (although I think you mean that git will only have `vX.Y` tags?)

RedbackThomson: Git will only have `rX.Y` branches, but all tags and releases will be `vX.Y.Z`

PR #52, opened 2019-12-06
Fixing HPO/BT deletion resource leak when SageMaker throttles Describe 

cadedaniel:
* The fix here is not clean; the delete portion of the code needs to be refactored.
* I added a unit test for each fix.
* I verified the fix by setting up nginx proxy to throttle 100% of requests, then I deleted the resources. Before my fix, they were leaked, after my change, the operator waits and deletes the Kubernetes resource after the SageMaker resource has been stopped.

### DIFF
--- a/controllers/batchtransformjob/batchtransformjob_controller.go
+++ b/controllers/batchtransformjob/batchtransformjob_controller.go
@@ -155,7 +155,7 @@ func (r *BatchTransformJobReconciler) reconcileJob(ctx reconcileRequestContext) 
 	} else {
 
 		ctx.Log.Info("Error getting batchtransformjob state in SageMaker", "requestErr", requestErr)
-		return r.handleSageMakerApiFailure(ctx, requestErr)
+		return r.handleSageMakerApiFailure(ctx, requestErr, false)
 	}
 
 }
@@ -180,12 +180,9 @@ func (r *BatchTransformJobReconciler) reconcileJobDeletion(ctx reconcileRequestC
 		} else {
 			// Case 2
 			log.Info("Sagemaker returns 4xx or 5xx or unrecoverable API Error")
-			if requestErr.StatusCode() == 400 {
-				// handleSageMakerAPIFailure does not removes the finalizer
-				r.removeFinalizerAndUpdate(ctx)
-			}
+
 			// Handle the 500 or unrecoverable API Error
-			return r.handleSageMakerApiFailure(ctx, requestErr)
+			return r.handleSageMakerApiFailure(ctx, requestErr, true)
 		}
 	} else {
 		log.Info("Job exists in Sagemaker, lets delete it")
@@ -227,7 +224,7 @@ func (r *BatchTransformJobReconciler) deleteBatchTransformJobIfFinalizerExists(c
 		_, err := req.Send(ctx)
 		if err != nil {
 			log.Error(err, "Unable to stop the job in sagemaker", "context", ctx)
-			return r.handleSageMakerApiFailure(ctx, err)
+			return r.handleSageMakerApiFailure(ctx, err, false)
 		}
 
 		return RequeueImmediately()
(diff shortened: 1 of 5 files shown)


RedbackThomson:
Do we have an existing ticket for refactoring the delete portion? Does this need to be a high priority, or is it likely this code will work for the foreseeable future?

controllers/controllertest/mock_sagemaker_client.go
```diff
@@ -274,9 +274,9 @@ func (m *MockSageMakerClientBuilder) AddDescribeEndpointErrorResponse(code strin
 }
 
 // Add a DescribeTrainingJob error response to the client.
-func (m *MockSageMakerClientBuilder) AddDescribeTransformJobErrorResponse(code string, statusCode int, reqId string) *MockSageMakerClientBuilder {
+func (m *MockSageMakerClientBuilder) AddDescribeTransformJobErrorResponse(code string, statusCode int, reqId, message string) *MockSageMakerClientBuilder {
 	m.responses.PushBack(describeTransformJobResponse{
-		err:  awserr.NewRequestFailure(awserr.New(code, "mock error message", fmt.Errorf(code)), statusCode, reqId),
+		err:  awserr.NewRequestFailure(awserr.New(code, message, fmt.Errorf(code)), statusCode, reqId),
```
RedbackThomson: Does this need to have a default value to not break existing tests?

cadedaniel: Nope, this is there because when I first wrote this I did not understand difference between code and message.

RedbackThomson: You'll have to explain to me the difference as well, then haha

controllers/batchtransformjob/batchtransformjob_controller.go
```diff
@@ -316,6 +313,11 @@ func (r *BatchTransformJobReconciler) handleSageMakerApiFailure(ctx reconcileReq
 			ctx.Log.Info("SageMaker rate limit exceeded, will retry", "err", awsErr)
```
gautamkmr: Are there other cases where we need to requeue ?

cadedaniel:
> Do we have an existing ticket for refactoring the delete portion? Does this need to be a high priority, or is it likely this code will work for the foreseeable future?

created https://issues.amazon.com/issues/P31245685 . This should be high priority from an ops perspective as the code is pretty brittle.

controllers/controllertest/mock_sagemaker_client.go
cadedaniel: Code is like a java classname, e.g. "ThrottleException", or "ValidationException". Message is the particular instance type, e.g. "No cross-account roles". SageMaker API does not use traditional HTTP codes (a lot of errors are multiplexed into HTTP 400), so we have to use both of these to extract the error cause.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -344,6 +340,11 @@ func (r *HyperparameterTuningJobReconciler) handleSageMakerApiFailure(ctx reconc
 			ctx.Log.Info("SageMaker rate limit exceeded, will retry", "err", awsErr)
 			return RequeueAfterInterval(r.PollInterval, nil)
 		} else if awsErr.StatusCode() == 400 {
+
+			if allowRemoveFinalizer {
```
gautamkmr: Which case will have `noRequeue` but removeFinalizer is `false`.

cadedaniel: Every case except the buggy case. Like I said in description, this fix is not pretty. https://issues.amazon.com/issues/P31245685 to clean up.

controllers/batchtransformjob/batchtransformjob_controller.go
cadedaniel: Yes, there are some cases where we should requeue but we do not. Any 4xx on [this page](https://docs.aws.amazon.com/sagemaker/latest/dg/CommonErrors.html) should be retried. I don't want to go about refactoring this without tests as it is a significant change. The comment on [this task](https://issues.amazon.com/issues/P31245685) explains that non-429 4xx errors are actually treated as 404s, which is a bug.

I would love to start working on that task this sprint, let's discuss in planning.

PR #53, opened 2019-12-07
Push smlogs binaries with tags

RedbackThomson:
## Description
Previously smlogs binaries were manually copied to new folders in the public buckets upon tagging. This change modifies the release script to copy binaries into the same folder hierarchy as the containers (`vX`, `vX.Y`, `vX.Y.Z`). 

This PR also refactors commonly used variables into `printf` format strings to ensure paths for binaries are in the same structure across multiple CodeBuild projects.


codebuild/scripts/deployment_variables.sh
```diff
@@ -0,0 +1,12 @@
+RELEASE_BUCKET_NAME_FMT="%s-%s"
```
cadedaniel: 🎉🎉🎉 love this file thx.

nit: Consider renaming to `deployment_constants` or `static_deployment_variables`

codebuild/scripts/release_tag.sh
```diff
@@ -57,5 +81,9 @@ for row in $(echo ${ACCOUNTS_ESCAPED} | jq -r '.[] | @base64'); do
   retag_image "$image" "$old_tag" "$minor_tag"
   retag_image "$image" "$old_tag" "$major_tag"
```
cadedaniel: Is `latest` tag automatically applied when pushing a new image? Will it be a problem if the corresponding smlogs binary is not tagged with `latest`?

RedbackThomson: Latest is always applied for both the images and the binaries upon running the deployment step. This step only handles the case where we want to apply a specific tag to that (or another) commit.

PR #54, opened 2019-12-09
Do not delete non-ephemeral cluster

gautamkmr:
Added instruction on not to delete non-ephemeral-cluster.

### DIFF
--- a/tests/codebuild/run_canarytest.sh
+++ b/tests/codebuild/run_canarytest.sh
@@ -36,9 +36,12 @@ function cleanup {
 
     delete_tests
 
-    # Tear down the cluster if we set it up.
-    echo "need_setup_cluster is true, tearing down cluster we created."
-    eksctl delete cluster --name "${cluster_name}" --region "${CLUSTER_REGION}"
+    if [ -z "${USE_EXISTING_CLUSTER}" ]
+    then
+        # Tear down the cluster if we set it up.
+        echo "USE_EXISTING_CLUSTER is true, tearing down cluster we created."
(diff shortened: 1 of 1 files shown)


tests/codebuild/run_canarytest.sh
```diff
@@ -36,9 +36,12 @@ function cleanup {
 
     delete_tests
 
-    # Tear down the cluster if we set it up.
-    echo "need_setup_cluster is true, tearing down cluster we created."
-    eksctl delete cluster --name "${cluster_name}" --region "${CLUSTER_REGION}"
+    if [ -z "${USE_EXISTING_CLUSTER}" ]
+    then
+        # Tear down the cluster if we set it up.
+        echo "need_setup_cluster is true, tearing down cluster we created."
```
cadedaniel: nit: Can you change log message to represent the condition?

tests/codebuild/run_canarytest.sh
```diff
@@ -36,9 +36,12 @@ function cleanup {
 
     delete_tests
 
-    # Tear down the cluster if we set it up.
-    echo "need_setup_cluster is true, tearing down cluster we created."
-    eksctl delete cluster --name "${cluster_name}" --region "${CLUSTER_REGION}"
+    if [ -z "${USE_EXISTING_CLUSTER}" ]
+    then
+        # Tear down the cluster if we set it up.
+        echo "USE_EXISTING_CLUSTER is true, tearing down cluster we created."
```
RedbackThomson: If USE_EXISTING_CLUSTER is FALSE we want to delete - not if it's true. Just the echo needs an update.

PR #60, opened 2019-12-19
Fixing batch transform

gautamkmr:
Added model creation along with spec.
Passed the dynamically generated sagemaker model
 to batch transform job.


gautamkmr:
@RedbackThomson

tests/codebuild/run_test.sh
```diff
@@ -6,6 +6,23 @@ function run_test()
...
+# This function gets the sagemaker model name from k8s model
+# Parameter:
+#    $1: Name of k8s model
+#    $2: Variable to store sagemaker model
+# e.g. get_sagemaker_model_from_k8s_model k8s_model sm_model
+#    This will populate the sm_model variable with sagemaker model.
+function get_sagemaker_model_from_k8s_model()
+{
+  local k8s_model_name="$1"
+  sagemaker_model_name="$2"
```
cadedaniel: I think there's a better way to share output. This doesn't require a variable to be initialized by caller (and no chance to have two different variable names).

When running the test:
```bash
sed -i "s/xgboost-model/$(get_sagemaker_model_from_k8s_model)/g" testfiles/xgboost-mnist-batchtransform.yaml
```

This function then would output the model name to stdout:
```bash
function get_sagemaker_model_from_k8s_model() {
   local k8s_model_name="$1"
   # stdout goes to caller
   kubectl get model $k8s_model_name | sed -n '2 p' |  awk '{print $3}' 
}
```

tests/codebuild/run_canarytest.sh
```diff
@@ -19,6 +19,7 @@ function delete_tests {
     kubectl delete hyperparametertuningjob --all
     kubectl delete trainingjob --all
     kubectl delete batchtransformjob --all
+    kubectl delete model --all
```
cadedaniel: nit: it's cleaner to delete models after hostingdeployments, otherwise controller will attempt to create the model again and then immediately delete it.

PR #70, opened 2020-01-07
Integration tests using EKS Fargate support

RedbackThomson:
Migrating integration tests to [EKS with AWS Fargate support](https://aws.amazon.com/blogs/aws/amazon-eks-on-aws-fargate-now-generally-available/) saves us at least 10 minutes*. Rather than waiting for EC2 nodes to be auto-scaled and then assigned to the cluster, we simply assign a Fargate profile to our operator namespace and let EKS do the rest.

* This number was taken from very basic preliminary tests


tests/codebuild/run_integtest.sh
```diff
@@ -73,10 +73,10 @@ set -e
 if [ "${need_setup_cluster}" == "true" ]; then
     echo "Launching the cluster"
     readonly cluster_name="sagemaker-k8s-pipeline-"$(date '+%Y-%m-%d-%H-%M-%S')""
-    readonly cluster_region="us-east-1"
```
cadedaniel: We should probably keep this line here until `cluster_region` is removed from the `need_setup_cluster != true` case. 

I only see [one (currently disabled) usage](https://github.com/aws/amazon-sagemaker-operator-for-k8s/search?q=cluster_region&unscoped_q=cluster_region) in integration tests, but the fact that this is disabled will be difficult to know for someone unfamiliar with the code.

RedbackThomson:
> I was under the impression that fargate setup would be faster because the fargate team keeps a warm pool of ec2 instances. Is the remaining startup time due to other cluster configuration?

From the simple testing that I did it seemed that creating a fargate profile (and especially deleting one) takes a fair amount of time. There must be some level of provisioning of resources happening in the background. Perhaps this will be decreased in the future once the product gains maturity.

tests/codebuild/run_integtest.sh
RedbackThomson: Sure. I will reassign this variable for use within integration tests. I didn't realise we were passing it around.

cadedaniel:
> > I was under the impression that fargate setup would be faster because the fargate team keeps
> > a warm pool of ec2 instances. Is the remaining startup time due to other cluster configuration?
> 
> From the simple testing that I did it seemed that creating a fargate profile (and especially deleting one) takes a fair amount of time. There must be some level of provisioning of resources happening in the background. Perhaps this will be decreased in the future once the product gains maturity.

Is it possible to save fargate profiles between runs? Do you think it's worth it to make a task to visit later?

RedbackThomson:
> Is it possible to save fargate profiles between runs? Do you think it's worth it to make a task to visit later?

Fargate profiles are only associated with a running cluster. They don't really exist as resources, more so they are a configuration inside EKS. Much like a ConfigMap they are applied to the cluster. 
I think this change already will significantly reduce our integration test times, so if we need to shave off more time later we can look into it more.

PR #71, opened 2020-01-08
Removing flakey wait from canaries

cadedaniel:
Canaries failed with the following:

```
+ kubectl wait --for=condition=Ready --timeout=5m 'pods/sagemaker-k8s-operator-controller-manager-9f488df7f-qjslx
sagemaker-k8s-operator-controller-manager-dfb5c5bb7-nhv57' -n sagemaker-k8s-operator-system
Error from server (NotFound): pods "sagemaker-k8s-operator-controller-manager-9f488df7f-qjslx\nsagemaker-k8s-operator-controller-manager-dfb5c5bb7-nhv57" not found
```

This is because the canaries have a wait command that used to be in the integration tests, but was removed for its flakiness.

### DIFF
--- a/tests/codebuild/run_canarytest.sh
+++ b/tests/codebuild/run_canarytest.sh
@@ -99,7 +99,7 @@ popd
 
 echo "Waiting for controller pod to be Ready"
 # Wait to increase chance that pod is ready
-# Should upgrade kubectl to version that supports `kubectl wait pods --all`
+# TODO: Should upgrade kubectl to version that supports `kubectl wait pods --all`
 sleep 60
 
 # Run the integration test file
--- a/tests/codebuild/run_canarytest.sh
+++ b/tests/codebuild/run_canarytest.sh
@@ -98,12 +98,9 @@ pushd sagemaker-k8s-operator
 popd 
 
 echo "Waiting for controller pod to be Ready"
-kubectl \
-    wait \
-    --for=condition=Ready \
-    --timeout=5m \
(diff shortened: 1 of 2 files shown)


tests/codebuild/run_canarytest.sh
```diff
@@ -98,12 +98,9 @@ pushd sagemaker-k8s-operator
...
 
 echo "Waiting for controller pod to be Ready"
-kubectl \
-    wait \
-    --for=condition=Ready \
-    --timeout=5m \
-    "pods/$(kubectl get pods -n sagemaker-k8s-operator-system | grep sagemaker-k8s-operator-controller-manager | awk '{print $1}')" \
-    -n sagemaker-k8s-operator-system 
+# Wait to increase chance that pod is ready
+# Should upgrade kubectl to version that supports `kubectl wait pods --all`
```
RedbackThomson: Lets just update `kubectl` then? We own the container that runs these tests

gautamkmr: Just make this as TODO:

cadedaniel: Agreed, this change is to bring the canary `wait` in line with the integration test `wait` to fix failures. I'll make a task for this.

RedbackThomson: How long would that take? If it's simply updating the version number inside a Dockerfile, it would be better to do that now and then change this PR to utilize the new CLI option rather than merge this?

cadedaniel: I am not confident that it will be that easy; I think there will have to be an existence check before waiting. I think this work should be prioritized against other optimization tasks for making tests faster.

PR #72, opened 2020-01-08
Remove the https reference from s3 paths

gautamkmr:
This came during one of the feedback from Arun Gupta 

"Change samples to use s3:// by default instead of https. This came up when he was changing regions, and asked if he had to change them for all https URLs." 


gautamkmr:
@cadedaniel  @RedbackThomson

RedbackThomson:
In our documentation do we have references to changing S3 URLs with `https://`?

gautamkmr:
> In our documentation do we have references to changing S3 URLs with `https://`?

I don't think so.

PR #73, opened 2020-01-09
Increase batch transform test timeouts

cadedaniel:
Canaries failed because BT took 23 minutes.

### DIFF
--- a/tests/codebuild/run_all_sample_canary_tests.sh
+++ b/tests/codebuild/run_all_sample_canary_tests.sh
@@ -28,7 +28,7 @@ run_test tests/xgboost-hosting-deployment.yaml
 # Format: `verify_test <type of job> <Job's metadata name> <timeout to complete the test> <desired status for job to achieve>` 
 verify_test TrainingJob xgboost-mnist 20m Completed
 verify_test HyperparameterTuningJob xgboost-mnist-hpo 20m Completed
-verify_test BatchTransformJob xgboost-batch 20m Completed 
+verify_test BatchTransformJob xgboost-batch 30m Completed
 verify_test HostingDeployment hosting 40m InService
 
 # Verify smlogs worked.
--- a/tests/codebuild/run_all_sample_test.sh
+++ b/tests/codebuild/run_all_sample_test.sh
@@ -53,7 +53,7 @@ verify_test trainingjob xgboost-mnist-custom-endpoint 20m Completed
 verify_test HyperparameterTuningJob xgboost-mnist-hpo 20m Completed
 verify_test HyperparameterTuningJob spot-xgboost-mnist-hpo 20m Completed
 verify_test HyperparameterTuningJob xgboost-mnist-hpo-custom-endpoint 20m Completed
-verify_test BatchTransformJob xgboost-batch 20m Completed
+verify_test BatchTransformJob xgboost-batch 30m Completed
 verify_test HostingDeployment hosting 40m InService
 
 # Verify smlogs worked.


cadedaniel:
@gautamkmr Doesn't it depend on how many rows we are transforming? In any case, I think metrics on canary job time would help us see how significant these outliers are.

RedbackThomson:
Can we change something within the spec to make this faster? Or use a different data set with fewer rows? 23 minutes for once test seems ridiculous

cadedaniel:
Good call... it would seem from the filename that this is only transforming 1000 instances `s3://sagemaker-sample-data-us-west-2/batch-transform/mnist-1000-samples`, which to me should take nowhere near 23 minutes. Perhaps we are using an instance type not designed for the workload? I will make a ticket to investigate and lower runtime here.

RedbackThomson:
> Good call... it would seem from the filename that this is only transforming 1000 instances `s3://sagemaker-sample-data-us-west-2/batch-transform/mnist-1000-samples`, which to me should take nowhere near 23 minutes. Perhaps we are using an instance type not designed for the workload? I will make a ticket to investigate and lower runtime here.

We are only running batch transform with 1 instance. Could we include in this PR an increase to 2 or 3? I would hope that divides the time up.

cadedaniel:
Yeah that should help.. I just have a hard time seeing how 1000 iterations of ~200ms (assumming a large model) should take that long. https://issues.amazon.com/issues/P31995373

RedbackThomson:
> Yeah that should help.. I just have a hard time seeing how 1000 iterations of ~200ms (assumming a large model) should take that long. https://issues.amazon.com/issues/P31995373

Bootstrapping initially and then writing to files/uploading to S3 would be my guess.

gautamkmr:
> Can we change something within the spec to make this faster? Or use a different data set with fewer rows? 23 minutes for once test seems ridiculous

did you mean job spec ? I think there is timeout section already exists but this time out is different.

cadedaniel:
Sorry I overlooked your suggestion `Could we include in this PR an increase to 2 or 3? I would hope that divides the time up.`. I think we should make task and prioritize it against other optimization tasks.. this one is probably pretty high though.

RedbackThomson:
> Sorry I overlooked your suggestion `Could we include in this PR an increase to 2 or 3? I would hope that divides the time up.`. I think we should make task and prioritize it against other optimization tasks.. this one is probably pretty high though.

I am already fixing integration tests to add a test case for deleting resources, I could also quickly check to see for performance improvements by increasing the number of instance in the spec.

cadedaniel:
> > Sorry I overlooked your suggestion Could we include in this PR an increase to 2 or 3? I would hope that divides the time up.. I think we should make task and prioritize it against other optimization tasks.. this one is probably pretty high though.
> 
> I am already fixing integration tests to add a test case for deleting resources, I could also quickly check to see for performance improvements by increasing the number of instance in the spec.

Feel free :). I think you should check first that the dataset actually has 1000 rows.. I don't like the idea of increasing instance counts ($$) unless we're sure we need to.

RedbackThomson:
> > > Sorry I overlooked your suggestion Could we include in this PR an increase to 2 or 3? I would hope that divides the time up.. I think we should make task and prioritize it against other optimization tasks.. this one is probably pretty high though.
> > 
> > 
> > I am already fixing integration tests to add a test case for deleting resources, I could also quickly check to see for performance improvements by increasing the number of instance in the spec.
> 
> Feel free :). I think you should check first that the dataset actually has 1000 rows.. I don't like the idea of increasing instance counts ($$) unless we're sure we need to.

SageMaker processing tasks are priced at instance type per hour. So hopefully if we can parallelise the data set into numerous instances, we would only pay a little extra for the amount of time spent bootstrapping that new instance.

cadedaniel:
@RedbackThomson I will merge this, if you're able to get 1/2 or 1/3 runtimes with increasing instance count then happy to revert, ok?

RedbackThomson:
> @RedbackThomson I will merge this, if you're able to get 1/2 or 1/3 runtimes with increasing instance count then happy to revert, ok?

Sure! Will know by EOD.

RedbackThomson:
> > @RedbackThomson I will merge this, if you're able to get 1/2 or 1/3 runtimes with increasing instance count then happy to revert, ok?
> 
> Sure! Will know by EOD.

Experimentally saw a reduction of down to less than 50% of the duration. I will include this change as part of my pull request tomorrow.

PR #74, opened 2020-01-11
Added integration tests for "delete" verb

RedbackThomson:
Added a new integration test function to check whether the "delete" verb reaches SageMaker and then successfully updates the local k8s resource state. Tested with TrainingJob, HyperparameterTuningJob and BatchTransformJob. Modifications will need to be made (a different method might need to be used) for HostingDeployments. 

This pull request also includes the following additions/fixes:
- Moved canary test file location to be in sync with integration test files
- Added a second instance to batch transform to decrease test times
- Refactored test scripts to use `yq` for YAML editing
- Removed redundant code


samples/xgboost-mnist-hpo-custom-endpoint.yaml
```diff
@@ -3,105 +3,105 @@ kind: HyperparameterTuningJob
...
-      enableInterContainerTrafficEncryption: false
+  region: us-west-2
+  sageMakerEndpoint: https://sagemaker.us-west-2.amazonaws.com
+  hyperParameterTuningJobConfig:
+    strategy: Bayesian
+    hyperParameterTuningJobObjective:
+      type: Minimize
+      metricName: validation:error
+    resourceLimits:
+      maxNumberOfTrainingJobs: 10
```
cadedaniel: Curious what motivation is behind this change?

tests/codebuild/run_test.sh
```diff
@@ -112,3 +112,54 @@ function build_fsx_from_s3()
...
+{
+    # Stop jobs so we can do PrivateLink test.
+    kubectl delete hyperparametertuningjob --all
+    kubectl delete trainingjob --all
+    kubectl delete batchtransformjob --all
+    kubectl delete hostingdeployment --all
+    kubectl delete model --all
+}
+
+# Delete a K8s resource and ensure all AWS resources were cleaned up successfully
```
cadedaniel: Can update to say that it starts and then deletes a k8s resource?

tests/codebuild/run_test.sh
```diff
@@ -112,3 +112,54 @@ function build_fsx_from_s3()
...
+       done' "$crd_type" "$job_name" "InProgress"
+   
+   if [ $? -ne 0 ]; then
+     echo "[FAILED] ${crd_type} ${job_name} job did not start within ${timeout}"
+     exit 1
+   fi
+
+   # Check that we can detect the status changing
+   kubectl delete -f "$file_name" &
+   timeout "${timeout}" bash -c \
```
cadedaniel: nit: this bit of logic is duplicated and can be pulled out into own function

tests/codebuild/run_test.sh
```diff
@@ -112,3 +112,54 @@ function build_fsx_from_s3()
...
+      'until [ "$(kubectl get "$0" "$1" -o=custom-columns=STATUS:.status | grep -i "$2" | wc -l)" -eq "1" ]; do \
+          sleep 5; \
+       done' "$crd_type" "$job_name" "Stopping"
+   
+   if [ $? -ne 0 ]; then
+     echo "[FAILED] ${crd_type} ${job_name} job did not start within ${timeout}"
```
cadedaniel: Update message (it says `did not start`)

tests/codebuild/run_test.sh
```diff
@@ -112,3 +112,54 @@ function build_fsx_from_s3()
...
+     exit 1
+   fi
+
+   echo "[PASSED] Verified ${crd_type} ${job_name} deleted successfully"
```
cadedaniel: So this function verifies that the SageMaker resource (as reported by k8s) is `Stopping`. Why do we not check against AWS proper? I want to make this a solid test that ignores any smart/buggy logic in the operator and confirms that the resources were deleted.

tests/codebuild/run_all_sample_test.sh
```diff
@@ -65,3 +64,15 @@ if [ "$(kubectl smlogs batchtransformjob xgboost-batch | wc -l)" -lt "1" ]; then
...
+delete_all_tests
+
+verify_delete TrainingJob testfiles/xgboost-mnist-trainingjob.yaml
+verify_delete HyperparameterTuningJob testfiles/xgboost-mnist-hpo.yaml
+
+# Create model before running batch delete test
+run_test testfiles/xgboost-model.yaml
+verify_test Model xgboost-model 1m Created
+yq w -i testfiles/xgboost-mnist-batchtransform.yaml "spec.modelName" "$(get_sagemaker_model_from_k8s_model xgboost-model)"
+verify_delete BatchTransformJob testfiles/xgboost-mnist-batchtransform.yaml 60s
```
cadedaniel: I'm amazed that this only takes 60 seconds / Training and HPO take 30 seconds. Should we put this higher to account for SageMaker p99s?

tests/codebuild/testfiles/spot-xgboost-mnist-hpo.yaml
```diff
@@ -3,19 +3,14 @@ kind: HyperparameterTuningJob
 metadata:
   name: spot-xgboost-mnist-hpo
 spec:
-    hyperParameterTuningJobConfig:
-        resourceLimits:
-            maxNumberOfTrainingJobs: 10
-            maxParallelTrainingJobs: 10
```
cadedaniel: Thanks for cleaning this up.

samples/xgboost-mnist-hpo-custom-endpoint.yaml
RedbackThomson: I fixed all the indentation and whitespace errors in all the integration tests and samples. Some tests also contained duplicate fields (not sure how Kubernetes handles redefining configuration fields) that needed to be removed.

tests/codebuild/run_all_sample_test.sh
RedbackThomson: It only barely starts any of the jobs before it tries to delete them. So the only lag is between the controller and SageMaker. It's likely these values will need to be adjusted, though.

tests/codebuild/run_test.sh
RedbackThomson: I initially started doing this but the messy part was extracting the job name from the spec considering they all have different JSON paths for their name. I wanted to make it a fairly compact test, but that wasn't really possible. My thought was that if we are detecting the state from SageMaker that our controller picking up a "Stopping" state would only happen if it picked it up from the SM resource directly.

samples/xgboost-mnist-hpo-custom-endpoint.yaml
RedbackThomson: Running 11 training jobs parallelising them 5 at a time means that you have 3 waves of training jobs. Decreasing this to 10 means you only run 2 waves. It was just a time save.

cadedaniel: awesome

tests/codebuild/run_test.sh
cadedaniel: I see. I mean this test is certainly better than what we had before.

> My thought was that if we are detecting the state from SageMaker that our controller picking up a "Stopping" state would only happen if it picked it up from the SM resource directly

This is true AFAIK, but we sometimes do smart things like reporting a certain state to make the k8s experience nice while doing things in the background. i.e. in HostingDeployment updates / deletes aren't supported when the SM endpoint is "Creating", so we lie and say "Deleting..." even though the actual status is "Creating". I realize you are not testing the HostingDeployment here; my concern is that future us might have to do something similar in one or more of the job controllers. At that point, this test is no longer useful in asserting that SageMaker resources are cleaned up. Does this make sense?

tests/codebuild/run_all_sample_test.sh
cadedaniel: ++ I think setting 60s for all of them would be great, left up to you though

tests/codebuild/run_all_sample_canary_tests.sh
```diff
@@ -42,3 +41,43 @@ if [ "$(kubectl smlogs batchtransformjob xgboost-batch | wc -l)" -lt "1" ]; then
...
+
+# Create model before running batch delete test
+run_test testfiles/xgboost-model.yaml
+verify_test Model xgboost-model 1m Created
+yq w -i testfiles/xgboost-mnist-batchtransform.yaml "spec.modelName" "$(get_sagemaker_model_from_k8s_model xgboost-model)"
+verify_delete BatchTransformJob testfiles/xgboost-mnist-batchtransform.yaml
+
+batch_job_name=$(kubectl get batchtransformjob xgboost-batch -o json | jq -r ".status.sageMakerTransformJobName")
+job_status=`aws sagemaker describe-transform-job --transform-job-name ${batch_job_name} --region us-west-2 | jq -r ".TransformJobStatus"`
+if [ "${job_status}" != "Stopping" ] && [ "${job_status}" != "Stopped" ]; then
```
gautamkmr: Does it make sense to put these lines into function? I understand these are different command but still some parts are common.

tests/codebuild/run_test.sh
```diff
@@ -57,19 +82,10 @@ function verify_test()
   fi
 
   echo "Waiting for job to complete"
-  timeout "${timeout}" bash -c \
-      'until [ "$(kubectl get "$0" "$1" -o=custom-columns=STATUS:.status | grep -i "$2" | wc -l)" -eq "1" ]; do \
-          echo "Job $1 has not completed yet"; \
-          sleep 5; \
-       done' "${crd_type}" "${crd_instance}" "${desired_status}"
+  wait_for_crd_status "$crd_type" "$crd_instance" "$timeout" "$desired_status"
```
gautamkmr: I liked this change.

tests/codebuild/run_integtest.sh
```diff
@@ -1,34 +1,23 @@
...
-}
-
 # A function to delete cluster, if cluster was not launched this will fail, so test will fail ultimately too
 function cleanup {
     # We want to run every command in this function, even if some fail.
     set +e
 
     echo "Controller manager logs:"
-    kubectl -n sagemaker-k8s-operator-system logs "$(kubectl get pods -n sagemaker-k8s-operator-system | grep sagemaker-k8s-operator-controller-manager | awk '{print $1}')" manager
+    # kubectl -n sagemaker-k8s-operator-system logs "$(kubectl get pods -n sagemaker-k8s-operator-system | grep sagemaker-k8s-operator-controller-manager | awk '{print $1}')" manager
```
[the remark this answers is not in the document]
RedbackThomson: Oops. I disabled this locally while debugging because it impossible in my console to scroll all the way back to when the error occurred.

tests/codebuild/run_all_sample_test.sh
```diff
@@ -65,3 +64,43 @@ if [ "$(kubectl smlogs batchtransformjob xgboost-batch | wc -l)" -lt "1" ]; then
...
     exit 1
 fi
+
+# Clean up resources before re-using metadata names
+delete_all_tests
+
+verify_delete TrainingJob testfiles/xgboost-mnist-trainingjob.yaml
+
+training_job_name=$(kubectl get trainingjob xgboost-mnist -o json | jq -r ".status.sageMakerTrainingJobName")
+job_status=`aws sagemaker describe-training-job --training-job-name ${training_job_name} --region us-west-2 | jq -r ".TrainingJobStatus"`
```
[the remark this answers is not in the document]
RedbackThomson: Wrapping in quotations will then require escaping the inner double quotations for variable names. I'm not sure how bash handles variable injection when escaping double quotations.

tests/codebuild/run_all_sample_canary_tests.sh
```diff
@@ -42,3 +41,43 @@ if [ "$(kubectl smlogs batchtransformjob xgboost-batch | wc -l)" -lt "1" ]; then
...
+  echo "[FAILED] AWS resource ${batch_job_name} did not stop (status \"${job_status}\")" 
+  exit 1
+else
+  echo "[PASSED] AWS resource ${batch_job_name} deleted from SageMaker" 
+fi
```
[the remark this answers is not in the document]
RedbackThomson: I'm in favour of moving all tests into their own subfolder - splitting them up by functionality. I was just following the pattern laid out previously, however I'm happy to change it now considering our tests are getting more verbose.

tests/codebuild/run_all_sample_canary_tests.sh
RedbackThomson: I initially put them into a function but that required something like 10 parameters and it became more confusing to understand how to call the function rather than just writing it out entirely.

tests/codebuild/run_test.sh
```diff
@@ -112,3 +112,54 @@ function build_fsx_from_s3()
...
+}
+
+# Delete a K8s resource and ensure all AWS resources were cleaned up successfully
+# Parameter:
+#    $1: CRD type
+#    $2: Filename of test
+#    $3: (Optional) Timeout
+# e.g. verify_delete trainingjob xgboost-mnist.yaml ".status.sageMakerTrainingJobName"
+function verify_delete()
+{
```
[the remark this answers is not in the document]
RedbackThomson: I wanted to ensure that when the training jobs are "running" that we can send them a "stop" signal when the user deletes the k8s resource. If we delete the resource after our tests have completed then all it is doing is deleting it from etcd - it does not make any API calls to SageMaker.

tests/codebuild/run_all_sample_test.sh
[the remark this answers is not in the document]
cadedaniel: I think you should try it out and see what happens :) Bash interprets double quotes differently than what we come to expect from other languages. In a nutshell, quoting expansions tells bash to treat it as a single term instead of splitting it on whitespace and processing special characters.

https://unix.stackexchange.com/a/131767/253226

tests/codebuild/run_test.sh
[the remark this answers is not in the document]
cadedaniel: I think the confusing thing here is the name of this function. It looks like all other tests we have (`verify_*`) yet performs the role of a helper function. If the name is changed to convey the fact that it is not a test alone and only a helper function, it will become much easier to grasp.

tests/codebuild/run_all_sample_canary_tests.sh
cadedaniel: A good way to do it is take only 2 parameters `crd_type crd_name` and use the CRD type parameter to determine the `jq -r ".status.*"` and `aws sagemaker describe-*-job` command syntax. I believe `Stopping` and `Stopped` is common for these three jobs, if not then that would have to be another parameter.

tests/codebuild/create_tests.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+source run_test.sh
+source inject_tests.sh
+
+function run_canary_tests
```
cadedaniel: Awesome refactor! Can you add a short documentation blurb to this file or the functions themselves? (one thing that would help is explaining how these differ from `delete_tests.sh`)

tests/codebuild/inject_tests.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+source run_test.sh
+
+function inject_all_variables
```
cadedaniel: ++ great refactor, can add comment so that file is self-explanatory? Why are we injecting variables, for example?

tests/codebuild/run_all_sample_test.sh
```diff
@@ -1,67 +1,12 @@
...
-    exit 1
-fi
-if [ "$(kubectl smlogs batchtransformjob xgboost-batch | wc -l)" -lt "1" ]; then
-    echo "smlogs batchtransformjob did not produce any output."
-    exit 1
-fi
+source create_tests.sh
+source delete_tests.sh
+source inject_tests.sh
+source smlogs_tests.sh
```
cadedaniel: 😍

tests/codebuild/run_test.sh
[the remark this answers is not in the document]
cadedaniel: maybe `setup_delete_test`?

tests/codebuild/delete_tests.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+
+source common.sh
+source create_tests.sh
```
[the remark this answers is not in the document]
RedbackThomson: It's just for that one instance of `verify_test`. That method is otherwise only ever used in `create_tests`, and it's really big. I just didn't want to move it out of that file ;(

tests/codebuild/delete_tests.sh
```diff
@@ -0,0 +1,122 @@
...
+  run_delete_canary_tests
+}
+
+# Applies a k8s resource, waits for it to start and then immediately deletes
+# and ensure the state of the resource in k8s is stopping.
+# Parameter:
+#    $1: CRD type
+#    $2: Filename of test
+#    $3: (Optional) Timeout
+# e.g. verify_delete trainingjob xgboost-mnist.yaml ".status.sageMakerTrainingJobName"
```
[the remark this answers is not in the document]
RedbackThomson: I looked into case insensitive string matching in bash and it's basically a nightmare unless we do blocks of if-else statements rather than a switch. 
I'll update the example and add a comment about proper casing.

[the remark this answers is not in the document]
cadedaniel: To do case-insensitive matching, I'd do this:
```bash
local crd_type="$(echo "$1" | awk '{print tolower($0)}')"
```

tests/codebuild/run_integtest.sh
```diff
@@ -1,34 +1,25 @@
...
     # We want to run every command in this function, even if some fail.
     set +e
 
-    echo "Controller manager logs:"
-    kubectl -n sagemaker-k8s-operator-system logs "$(kubectl get pods -n sagemaker-k8s-operator-system | grep sagemaker-k8s-operator-controller-manager | awk '{print $1}')" manager
-
-    # Describe, if the test fails the Additional field might have more helpful info.
-    echo "trainingjob description:"
-    kubectl describe trainingjob
+    if [ "${PRINT_DEBUG}" != "false" ]; then
```
cadedaniel: nit: If we use `PRINT_DEBUG` as name for this flag, then our canaries and integration tests will always run in "debug mode" which feels weird. Consider `CI_MODE`?

RedbackThomson: This only checks to see if PRINT_DEBUG!="false" to print the logs. That means the default usage is that it prints these logs. I only put this statement in so that a developer has the option to disable it - but they have to explicitly set that one variable.

I think a move to "developer mode" or the converse "ci mode"/"verbose mode" would be preferred for all defaults but it's outside the scope of this already bloated PR lol

PR #77, opened 2020-01-17
Adding generate_trust_policy script to make generation of trust.json …

cadedaniel:
…easier

I will link to this in a documentation change.

https://issues.amazon.com/issues/P31970283

### DIFF
--- a/scripts/generate_trust_policy
+++ b/scripts/generate_trust_policy
@@ -4,16 +4,18 @@
 # https://sagemaker.readthedocs.io/en/stable/amazon_sagemaker_operators_for_kubernetes.html#create-an-iam-role
 #
 # Run as:
-# $ ./generate_trust_policy ${EKS_CLUSTER_REGION} ${AWS_ACCOUNT_ID} ${OIDC_ID} > trust.json
+# $ ./generate_trust_policy ${EKS_CLUSTER_REGION} ${AWS_ACCOUNT_ID} ${OIDC_ID} [optional: ${OPERATOR_NAMESPACE}] > trust.json
 #
 # For example:
 # $ ./generate_trust_policy us-west-2 123456789012 D48675832CA65BD10A532F597OIDCID > trust.json
 # This will create a file `trust.json` containing a role policy that enables the operator in an EKS cluster to assume AWS roles.
 #
+# The OPERATOR_NAMESPACE parameter is for when you want to run the operator in a custom namespace other than "sagemaker-k8s-operator-system".
 
 cluster_region="$1"
 account_number="$2"
 oidc_id="$3"
+operator_namespace="${4:-sagemaker-k8s-operator-system}"
 
 printf '{
   "Version": "2012-10-17",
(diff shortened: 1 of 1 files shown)


scripts/generate_trust_policy
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
RedbackThomson: Can we use `/bin/bash` to match up with the other shell files?

scripts/generate_trust_policy
```diff
@@ -0,0 +1,36 @@
...
+# Run as:
+# $ ./generate_trust_policy ${EKS_CLUSTER_REGION} ${AWS_ACCOUNT_ID} ${OIDC_ID} > trust.json
+#
+# For example:
+# $ ./generate_trust_policy us-west-2 123456789012 D48675832CA65BD10A532F597OIDCID > trust.json
+# This will create a file `trust.json` containing a role policy that enables the operator in an EKS cluster to assume AWS roles.
+#
+
+cluster_region="$1"
+account_number="$2"
```
RedbackThomson: We could also generate this from AWSCLI using `aws sts get-caller-identity | jq -r ".Account"`. Most people probably do not have their account number on hand, so this might otherwise require more documentation.

scripts/generate_trust_policy
cadedaniel: `env bash` is safer for larger distribution as some distros place bash in different places. The other ones should be using `env bash` but it's not as impactful since we control the environment.

https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

scripts/generate_trust_policy
cadedaniel: I don't know.. while I agree that is useful, I think this script is more valuable as a dumb script. For example if someone has multiple AWS accounts and are setting it up for one but `aws` picks another. I think our users will be able to get their account number.

scripts/generate_trust_policy
```diff
@@ -0,0 +1,36 @@
...
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Federated": "arn:aws:iam::'"${account_number}"':oidc-provider/oidc.eks.'"${cluster_region}"'.amazonaws.com/id/'"${oidc_id}"'"
+      },
+      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Condition": {
+        "StringEquals": {
+          "oidc.eks.'"${cluster_region}"'.amazonaws.com/id/'"${oidc_id}"':aud": "sts.amazonaws.com",
+          "oidc.eks.'"${cluster_region}"'.amazonaws.com/id/'"${oidc_id}"':sub": "system:serviceaccount:sagemaker-k8s-operator-system:sagemaker-k8s-operator-default"
```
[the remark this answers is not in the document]
cadedaniel: Good call out, will add as optional parameter.

scripts/generate_trust_policy
cadedaniel: Sorry I missed the overall comment on the point + justification.

I like your justification and think there's merit.. I'll make this script smarter so that it grabs everything.

cadedaniel:
Closing until this is prioritized again.

PR #81, opened 2020-01-22
Fix canary missing scripts

RedbackThomson:
The canary container Dockerfile was using old references to shell scripts that were renamed in the last refactor. This adds all scripts back into the container.


RedbackThomson:
> Were you able to test this?

Tested it locally, but didn't want to wait for EKSCTL. However it properly sourced the test scripts.

PR #85, opened 2020-01-29
Fix unit testing lint step

RedbackThomson:
Install the golint tool (if necessary) before running linting task. This initial step has no impact once it has been installed. A linting failure does not return a non-zero exit code.


Makefile
```diff
@@ -6,15 +6,15 @@ CRD_OPTIONS ?= "crd:trivialVersions=true"
 all: manager
 
 # Run tests
-test: generate fmt vet manifests
+test: lint generate fmt vet manifests
```
cadedaniel: What is your logic for making `lint` a separate `make` target?

RedbackThomson: Contributors are able to run `make lint` before submitting code reviews.

cadedaniel: Why not make `vet` have `lint` as a target, so that all of these other target dependencies are more consolidated?

PR #86, opened 2020-01-30
Refactored HyperParameterTuningJob controller 

RedbackThomson:
- Updated the HyperParameterTuningJob controller to match the HostingDeployment and TrainingJob reconciliation flow.
- Modified the HPO Training Job spawner to use the new SageMaker client wrapper.
- Updated all unit tests to meet newest specification (based off HostingDeployment).

Tested by running to completion and by stopping during InProgress the integration test HPO job.


controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
+		}
+		break
 
-func (r *HyperparameterTuningJobReconciler) getHyperParameterTuningJobName(state hpojobv1.HyperparameterTuningJob) string {
-	return GetGeneratedJobName(state.ObjectMeta.GetUID(), state.ObjectMeta.GetName(), 32)
-}
+	case sagemaker.HyperParameterTuningJobStatusCompleted:
+		if err = r.addBestTrainingJobToStatus(ctx); err != nil {
+			return r.updateStatusAndReturnError(ctx, ReconcilingTuningJobStatus, errors.Wrap(err, "Unable to add best training job to status"))
+		}
```
cadedaniel: This should happen for all statuses. The best training job can even be the first training job to complete -- it is updated over the lifetime of the tuning job.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
-// Extract the BestTrainingJob from the SageMaker description if it exists and convert it to our Kubernetes type.
-func (r *HyperparameterTuningJobReconciler) createBestTrainingJob(ctx reconcileRequestContext) *commonv1.HyperParameterTrainingJobSummary {
-
-	if ctx.SageMakerDescription.BestTrainingJob == nil {
-		ctx.Log.Info("Cannot get BestTrainingJob: No BestTrainingJob in HPO description")
-		return nil
+// Clean up all models and endpoints, then remove the HostingDeployment finalizer.
+func (r *Reconciler) addBestTrainingJobToStatus(ctx reconcileRequestContext) error {
+	if ctx.TuningJobDescription.BestTrainingJob == nil {
+		return errors.New("No BestTrainingJob in HPO description")
```
cadedaniel: It seems like a bad idea to fail hard here. The BestTrainingJob will not always be present.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
+		return err
 	}
 
-	return bestTrainingJob
+	ctx.TuningJob.Status.BestTrainingJob = bestTrainingJob
+
+	return nil
 }
 
 // Convert an aws-go-sdk-v2 HyperParameterTrainingJobSummary to a Kubernetes SageMaker type, returning errors if there are any.
```
cadedaniel: Can we move this function to SDK converters?

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
+	return nil
 }
 
-// Extract the BestTrainingJob from the SageMaker description if it exists and convert it to our Kubernetes type.
-func (r *HyperparameterTuningJobReconciler) createBestTrainingJob(ctx reconcileRequestContext) *commonv1.HyperParameterTrainingJobSummary {
-
-	if ctx.SageMakerDescription.BestTrainingJob == nil {
-		ctx.Log.Info("Cannot get BestTrainingJob: No BestTrainingJob in HPO description")
-		return nil
+// Clean up all models and endpoints, then remove the HostingDeployment finalizer.
```
cadedaniel: UPdate comment

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
 
 	return nil
 }
 
-func (r *HyperparameterTuningJobReconciler) statusMatchesDescription(ctx reconcileRequestContext) bool {
-	// TODO implement
-	return false
+// If job is running, stop it and requeue
+// If job is stopping, update status and requeue
+// If job has finished, failed or stopped remove finalizer and don't requeue
```
cadedaniel: Update comments please.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
+func (r *Reconciler) deleteHyperparameterTuningJob(ctx reconcileRequestContext) error {
```
cadedaniel: Where is this code called?

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
-	if awsErr, requestFailed := describeError.(awserr.RequestFailure); requestFailed {
-		if r.isSageMaker404Response(awsErr) {
-			log.Info("Job does not exist in sagemaker")
-			return nil, nil
-		} else {
-			log.Info("Non-404 error response from DescribeHyperparameterTuningJob")
-			return nil, awsErr
+	case sagemaker.HyperParameterTuningJobStatusStopped, sagemaker.HyperParameterTuningJobStatusFailed:
+		if controllers.HasDeletionTimestamp(ctx.TuningJob.ObjectMeta) {
+			return r.cleanupAndRemoveFinalizer(ctx)
```
cadedaniel: Why is this duplicated? Line 168 does this, this should noop and let 168 handle cleanup.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -488,17 +373,6 @@ func (r *HyperparameterTuningJobReconciler) SetupWithManager(mgr ctrl.Manager) e
...
-	return awsError.Code() == hpoResourceNotFoundApiCode
-}
-
-// When we run describeHPOJob with the name of job, sagemaker returns throttling exception
-// with error code 400 instead of 429.
-func (r *HyperparameterTuningJobReconciler) isSageMaker429Response(awsError awserr.RequestFailure) bool {
-	return (awsError.Code() == "ThrottlingException") && (awsError.Message() == "Rate exceeded")
-}
-
 func newTrainingJobStatusCountersFromDescription(sageMakerDescription *sagemaker.DescribeHyperParameterTuningJobOutput) *commonv1.TrainingJobStatusCounters {
```
cadedaniel: This code is no longer called, but should be.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
```
cadedaniel: Reminder for me to look at unit tests in more depth on second pass.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
+					})
+				})
+			})
+
+			When("HasDeletionTimestamp", func() {
+				BeforeEach(func() {
+					shouldHaveDeletionTimestamp = true
+					expectedStatus = sagemaker.HyperParameterTuningJobStatusStopping
+					mockSageMakerClientBuilder.
+						AddStopHyperParameterTuningJobResponse(sagemaker.StopHyperParam
```
cadedaniel: Update comment, it no longer does nothing.

cadedaniel: IMO `callTrackingHPOTrainingJobSpawner` tells readers more about what this type does. Consider renaming to something more descriptive than "noop"

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -1701,34 +559,109 @@ func createHpoJob(withFinalizer bool, k8sName, k8sNamespace, sageMakerName, regi
...
+func ExpectRequestToStopHyperParameterTuningJob(req interface{}, tuningJob *hpojobv1.HyperparameterTuningJob) {
+	Expect(req).To(BeAssignableToTypeOf((*sagemaker.StopHyperParameterTuningJobInput)(nil)))
+
+	stopRequest := req.(*sagemaker.StopHyperParameterTuningJobInput)
+	Expect(*stopRequest.HyperParameterTuningJobName).To(Equal(controllers.GetGeneratedJobName(tuningJob.ObjectMeta.GetUID(), tuningJob.ObjectMeta.GetName(), MaxHyperParameterTuningJobNameLength)))
+}
+
+// Helper function to verify that the controller attempted to delete the spawned training jobs.
+func ExpectDeletedSpawnedTrainingJobs(spawner noopHPOTrainingJobSpawner) {
+	Expect(spawner.deleteSpawnedTrainingJobsCalls).ToNot(Equal(0))
```
cadedaniel: Why not make test more precise and verify that it is called once?

smlogs-kubectl-plugin/pkg/cmd/smlogs_test.go
```diff
@@ -25,12 +25,12 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/aws/awserr"
-	cwlogs "github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs"
 	transformjobv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/batchtransformjob"
 	commonv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/common"
 	trainingjobv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/trainingjob"
+	"github.com/aws/aws-sdk-go-v2/aws"
```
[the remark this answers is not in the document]
RedbackThomson: This wasn't done by me intentionally. I wonder if the linter in my editor switched the order.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
RedbackThomson: Line 168 only acts if there is no description available. This handles the case where we have already successfully stopped the SageMaker resource. They are both cleaned up in the same way, however.

controllers/sdkutil/spec_comparison.go
```diff
@@ -39,7 +39,7 @@ type Comparison struct {
 // Determine if the given TrainingJobSpec matches the DescribeTrainingJobOutput. This converts the description to a TrainingJobSpec,
 // then selectively compares fields.
 func TrainingJobSpecMatchesDescription(description sagemaker.DescribeTrainingJobOutput, spec trainingjobv1.TrainingJobSpec) Comparison {
-	remoteSpec := CreateTrainingJobSpecFromDescription(description)
+	remoteSpec, _ := CreateTrainingJobSpecFromDescription(description)
```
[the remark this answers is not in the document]
RedbackThomson: Every field specified in the spec will be different from that specified in the description (no spec will match any description). This `TrainingJobSpecMatchesDescription` doesn't seem to be used anywhere within the codebase anymore. I suppose it used to work with TrainingJob to reconcile the two, but no longer is necessary?

controllers/hyperparametertuningjob/hpo_trainingjob_spawner_test.go
```diff
@@ -517,25 +535,10 @@ var _ = Describe("DeleteSpawnedTrainingJobs", func() {
...
-		spawner = createHpoTrainingJobSpawner(k8sClient, logf.Log, sageMakerClient)
-
-		// Run test
-		err = spawner.DeleteSpawnedTrainingJobs(context.Background(), *createHpoJob(false, "hpo-k8s-name", namespace, "hpo-sagemaker-name", "aws-region"))
-
-		// Verify expectations
-		Expect(err).To(HaveOccurred())
-	})
-
-	It("Gracefully handles k8s Get failure", func() {
```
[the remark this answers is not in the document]
RedbackThomson: I deleted it because if the spawned training job has already been deleted, we would not want to expect an error to be returned. However, now that we want to check 404's separately from other Get errors, it might be better to add it back in.

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
```diff
@@ -291,33 +292,22 @@ func (s hpoTrainingJobSpawner) deleteSpawnedTrainingJobsConcurrently(ctx context
...
 		}
 	}
 
 	if needsDelete {
-		s.Log.Info("Deleting TrainingJob", "trainingJobName", trainingJobName)
-		if err := s.K8sClient.Delete(ctx, &trainingJob); err != nil {
-			return IgnoreNotFound(err)
+		s.Log.Info("Deleting TrainingJob", "trainingJobName", trainingJob.Status.SageMakerTrainingJobName)
+		if err := s.K8sClient.Delete(ctx, trainingJob); err != nil {
+			return errors.Wrap(err, "Failed to delete training job")
```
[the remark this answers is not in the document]
RedbackThomson: This method should only be called after a successful `Get`. I don't want to continue checking the same error conditions within each method.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -19,437 +19,292 @@ package hyperparametertuningjob
...
+	return nil
+}
 
-	if _, createError := request.Send(ctx); createError == nil {
-		ctx.Log.Info("HyperParameterTuningJob created in SageMaker")
-		return RequeueImmediately()
+// Initialize fields on the context object which will be used later.
+func (r *Reconciler) initializeContext(ctx *reconcileRequestContext) error {
+	// Ensure we are using the job name specified in the spec
+	if ctx.TuningJob.Spec.HyperParameterTuningJobName != nil {
```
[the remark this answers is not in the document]
RedbackThomson: We decided that it would be better to do input validation (minimum length requirements) in the CRD spec, instead.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
+					})
+				})
+			})
+
+			When("HasDeletionTimestamp", func() {
+				BeforeEach(func() {
+					shouldHaveDeletionTimestamp = true
+					expectedStatus = sagemaker.HyperParameterTuningJobStatusStopping
+					mockSageMakerClientBuilder.
+						AddStopHyperParameterTuningJ
```
cadedaniel: We should have a test for when the delete of spawned training job fails. The controller needs to retry, otherwise it will leave a bunch of jobs around that have to be deleted manually (i.e. a special finalizer removed, then deleted).

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
-
-		// Verify that the hpoTrainingJobSpawner was called
-		Expect(spawnMissingTrainingJobHpoJobs.Len()).To(Equal(1))
+	Context("HyperParameterTuningJob exists", func() {
 
-		// Verify that the hpoTrainingJobSpawner received the correct parameters.
-		actualHpoJob := spawnMissingTrainingJobHpoJobs.Front().Value.(hpojobv1.HyperparameterTuningJob)
-		Expect(*actualHpoJob.Spec.HyperParameterTuningJobName).To(Equal(*job.Spec.HyperParameterTuningJobName))
-		Expect(*actualHpoJob.Spec.Region).To(Equal(*job.Spec.Region))
-		Expect(actualHpoJob.ObjectMeta.GetNamespace()).To(Equal(job.ObjectMeta.GetNamespace()))
```
cadedaniel: Why are we removing these assertions? We should verify that we invoke our mocked components correctly.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
-			Namespace: job.ObjectMeta.Namespace,
-			Name:      job.ObjectMeta.Name,
-		}, job)
-		Expect(err).ToNot(HaveOccurred())
+				req := receivedRequests.Front().Next().Value
+				Expect(req).To(BeAssignableToTypeOf((*sagemaker.CreateHyperParameterTuningJobInput)(nil)))
 
-		Expect(job.Status.TrainingJobStatusCounters).ToNot(BeNil())
+				createdRequest := req.(*sagemaker.CreateHyperParameterTuningJobInput)
+				Expect(*createdRequest.HyperParameterTuningJobName).To(Equal(controllers.GetGeneratedJobName(tuningJob.ObjectMeta.GetUID(), tuningJob.ObjectMeta.GetName(), MaxHyperParameterTuningJobNameLength)))
```
cadedaniel: This code is duplicated, why not pull out into function (like `ExpectRequestToStopHyperParameterTuningJob`)?

cadedaniel: This will improve readability as well.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
+			})
+		})
 
-		// Create controller with mocks and its request
-		controller := createHpoReconciler(k8sClient, CreateMockSageMakerClientProvider(sageMakerClient), createMockHpoTrainingJobSpawnerProvider(noopHpoTrainingJobSpawner{}), CreateMockAwsConfigLoader(), "1s")
-		request := CreateReconciliationRequest(job.ObjectMeta.Name, job.ObjectMeta.Namespace)
+		Context("!HasDeletionTimestamp", func() {
+			BeforeEach(func() {
+				mockSageMakerClientBuilder.
+					AddCreateHyperParameterTuningJobResponse(sagemaker.CreateHyperParameterTuningJobOutput{}).
```
cadedaniel: We should have tests that assert any failure to create is propagated to the user via the `Status.Additional` field. This is particularly important because any s3 region error, iam error, capacity error, et al will surface to customer this way.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
+					AddDescribeHyperParameterTuningJobResponse(CreateDescribeOutputWithOnlyStatus(expectedStatus))
+			})
+
+			When("!HasDeletionTimestamp", func() {
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+
+				It("Updates status", func() {
+					ExpectStatusToBe(tuningJob, string(expectedStatus))
```
cadedaniel: Man, generics would be so powerful here. I wonder if there's a way we can rig our own generics by iterating over a set of preconditions that should match some assumption. It would look more like Go table-tests but would likely reduce how many tests we duplicate.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
cadedaniel: remove secondary status artifact

cadedaniel: We should have a test that asserts that the BestTrainingJob is set, when present.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
-		var stopped int64 = 5
-
-		// Set mock error values.
-		description.TrainingJobStatusCounters = &sagemaker.TrainingJobStatusCounters{
-			Completed:         &completed,
-			InProgress:        &inProgress,
-			NonRetryableError: &nonRetryableError,
-			RetryableError:    &retryableError,
-			Stopped:           &stopped,
-		}
```
cadedaniel: Why are we no longer testing this?

controllers/sdkutil/spec_sdk_converters.go
```diff
@@ -195,6 +175,56 @@ func createCreateHyperParameterTuningJobInputFromSpec(spec hpojobv1.Hyperparamet
...
+	}
+
+	json.Unmarshal(str, &target)
+
+	target.TunedHyperParameters = tunedHyperParameters
+	return &target, nil
+}
+
+// CreateTrainingJobStatusCountersFromDescription creates a set of TrainingJobStatusCounters from a DescribeHyperParameterTuningJobOutput
+func CreateTrainingJobStatusCountersFromDescription(sageMakerDescription *sagemaker.DescribeHyperParameterTuningJobOutput) *commonv1.TrainingJobStatusCounters {
```
cadedaniel: ++, thanks for moving this here.

controllers/sdkutil/spec_comparison.go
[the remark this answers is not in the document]
RedbackThomson: Should I simply delete this method if it is never being called? Why TODO or simply comment it?

In regards to my previous comment, what I meant to say was that if the method returns a (nil, error) tuple, the comparison returned by this method will show that no field in the description (the nil object) matches the training job.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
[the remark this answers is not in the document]
RedbackThomson: I think all the CRD spec changes should be their own PR since there are probably other fields we should validate have non-empty validation.

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
```diff
@@ -239,22 +226,33 @@ func (s hpoTrainingJobSpawner) deleteSpawnedTrainingJobsConcurrently(ctx context
...
+
+				key := types.NamespacedName{
+					Namespace: k8sNamespace,
+					Name:      *trainingJobSummary.TrainingJobName,
+				}
+
+				var trainingJob trainingjobv1.TrainingJob
+				if err := s.K8sClient.Get(ctx, key, &trainingJob); err != nil {
+					// If the job has previously been deleted, we are done with it
+					if apierrs.IsNotFound(err) {
```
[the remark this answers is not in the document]
RedbackThomson: I don't see how I could refactor this to reduce complexity. There are 3 different code paths that each have separate conditions. What did you have in mind?

controllers/sdkutil/clientwrapper/sagemaker_client.go
```diff
@@ -414,3 +499,42 @@ func IsStopTrainingJob404Error(err error) bool {
...
+
+// HyperParameterTuningJobPaginator wraps the SageMaker ListTrainingJobsForHyperParameterTuningJobPaginator.
+type HyperParameterTuningJobPaginator interface {
+	Next(context.Context) bool
+	CurrentPage() []sagemaker.HyperParameterTrainingJobSummary
+	Err() error
+}
+
+type hyperParameterTuningJobPaginator struct {
+	HyperParameterTuningJobPaginator
```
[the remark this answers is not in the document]
RedbackThomson: I saw you doing this in other areas of the code but it was being inconsistently used. From now on I will use this pattern, though.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
-		}, job)
-		if err != nil && apierrs.IsNotFound(err) {
-			return
-		}
-		Expect(err).ToNot(HaveOccurred())
-
-		// Assert that deletionTimestamp is nonzero.
-		err = k8sClient.Delete(context.Background(), job)
-		Expect(err).ToNot(HaveOccurred())
+	It("should not requeue if the HyperParameterTuningJob does not exist", func() {
```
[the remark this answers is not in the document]
RedbackThomson: I simply copied them from HostingDeployment, but they look much prettier when using `BeforeEach`.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
RedbackThomson: This seems like a unit test addition for every controller. Perhaps lets break this into another refactor task?

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
RedbackThomson: Pull what out into which function?

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,538 @@ package hyperparametertuningjob
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	// +kubebuilder:scaffold:imports
+)
 
-		// Run test
-		controller.Reconcile(request)
-		Expect(actualRegion).To(Equal(expectedRegion))
-	})
-})
+var _ = Describe("Reconciling a HyperParameterTuningJob while failing to get the Kubernetes job", func() {
```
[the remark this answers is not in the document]
RedbackThomson: After all of these improvements, 80.3%

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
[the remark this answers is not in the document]
RedbackThomson: I imagine I can settle this by adding `&& !apierrs.IsNotFound(err)` to the conditional? Is that the condition you were worried about occurring?

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
RedbackThomson: I see what you mean. The paginator for spawned training jobs would return a 404 if we attempt to clean up a job that's not in SageMaker. Good catch, thank you.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -18,1680 +18,645 @@ package hyperparametertuningjob
...
+		It("should not requeue", func() {
+			ExpectNoRequeue(reconcileResult, reconcileError)
+		})
 	})
 
-	It("should update status and not requeue if error is 4xx", func() {
+	Context("An error occurred with the K8s client", func() {
+		BeforeEach(func() {
+			mockK8sClient := FailToGetK8sClient{}
+			reconciler = createReconciler(mockK8sClient, sageMakerClient, "1s", mockTrackingHPOTrainingJobSpawner{})
```
cadedaniel: nit: can deduplicate this code by providing an overwritable k8sClient for the `BeforeEach` functions, and move this line to `JustBeforeEach`. See the other tests (line 118 in this file) for how I did it.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
cadedaniel: I refer to the following lines:

```go
req := receivedRequests.Front().Next().Value
Expect(req).To(BeAssignableToTypeOf((*sagemaker.CreateHyperParameterTuningJobInput)(nil)))

createdRequest := req.(*sagemaker.CreateHyperParameterTuningJobInput)
Expect(*createdRequest.HyperParameterTuningJobName).To(Equal(controllers.GetGeneratedJobName(tuningJob.ObjectMeta.GetUID(), tuningJob.ObjectMeta.GetName(), MaxHyperParameterTuningJobNameLength)))
```

```go
// from "Spec defines HyperParameterTuningJobName"
req := receivedRequests.Front().Next().Value
Expect(req).To(BeAssignableToTypeOf((*sagemaker.CreateHyperParameterTuningJobInput)(nil)))

createdRequest := req.(*sagemaker.CreateHyperParameterTuningJobInput)
Expect(*createdRequest.HyperParameterTuningJobName).To(Equal("tuning-job-name"))
```

They are identical in everything except the expected job name.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
cadedaniel: ++, I'll add this comment to https://issues.amazon.com/issues/P31113513

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -1701,34 +666,140 @@ func createHpoJob(withFinalizer bool, k8sName, k8sNamespace, sageMakerName, regi
...
+func ExpectRequestToStopHyperParameterTuningJob(req interface{}, tuningJob *hpojobv1.HyperparameterTuningJob) {
+	Expect(req).To(BeAssignableToTypeOf((*sagemaker.StopHyperParameterTuningJobInput)(nil)))
+
+	stopRequest := req.(*sagemaker.StopHyperParameterTuningJobInput)
+	Expect(*stopRequest.HyperParameterTuningJobName).To(Equal(controllers.GetGeneratedJobName(tuningJob.ObjectMeta.GetUID(), tuningJob.ObjectMeta.GetName(), MaxHyperParameterTuningJobNameLength)))
+}
+
+// Helper function to verify that the controller attempted to delete the spawned training jobs.
+func ExpectDeletedSpawnedTrainingJobs(spawner mockTrackingHPOTrainingJobSpawner, calls int) {
+	Expect(spawner.deleteSpawnedTrainingJobsCalls.Len()).To(Equal(calls))
```
cadedaniel: We shouldn't reduce the quality of these tests -- the unit tests before checked that we called the spawner deletion correctly. Can you re-add the tests for this, just like you did with `ExpectSpawnMissingTrainingJobs`?

controllers/sdkutil/spec_comparison.go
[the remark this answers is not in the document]
cadedaniel: > Should I simply delete this method if it is never being called? Why TODO or simply comment it?

I like to think holistically when making changes to reduce rough edges when the change is implemented. If we have a TODO, when we revisit this in the JSON refactor, we can make sure to remove this completely. I'm not opposed to removing since it is dead code, we just need to be vigilant in JSON refactor to not leave any rough edges.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -41,34 +41,48 @@ import (
...
 	)
 
 	BeforeEach(func() {
 		sageMakerClient = NewMockSageMakerClientBuilder(GinkgoT()).Build()
 	})
 
-	It("should not requeue if the TrainingJob does not exist", func() {
-		controller := createReconcilerWithMockedDependencies(k8sClient, sageMakerClient, "1s")
-
+	JustBeforeEach(func() {
```
cadedaniel: thanks for improving this :)

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -18,505 +18,339 @@ package hyperparametertuningjob
...
+	return nil
 }
 
-func (r *HyperparameterTuningJobReconciler) getHyperParameterTuningJobName(state hpojobv1.HyperparameterTuningJob) string {
-	return GetGeneratedJobName(state.ObjectMeta.GetUID(), state.ObjectMeta.GetName(), 32)
-}
+// Initialize fields on the context object which will be used later.
+func (r *Reconciler) initializeContext(ctx *reconcileRequestContext) error {
+	// Ensure we generate a new name and populate the spec if none was specified
+	if ctx.TuningJob.Spec.HyperParameterTuningJobName == nil {
```
cadedaniel: I can't respond to the earlier thread due to either my lack of understanding of how to use GitHub PR or GitHub PR has a bug.

> @cadedaniel
> cadedaniel 7 days ago Member
> 
> What if the name is empty but not nil?
> 
>  @cadedaniel
> cadedaniel 7 days ago Member
> This might be a bug in TrainingJob as well.
> 
>  @RedbackThomson
> RedbackThomson 3 days ago Author Member
> We decided that it would be better to do input validation (minimum length requirements) in the CRD spec, instead.
> 
>  @cadedaniel
> cadedaniel 3 days ago Member
> I don't see minimum length validation for HPO or TJ name. Can you include change for HPO and fix TJ?
> 
>  @RedbackThomson
> RedbackThomson 2 days ago Author Member
> I think all the CRD spec changes should be their own PR since there are probably other fields we should validate have non-empty validation.

We should improve validation in another PR, but we can't make the controller rely on validation that doesn't exist. The fix here is just adding MinLength validation to HPO and TJ. If you don't want to expand scope by adding validation, then this line should check for empty name.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller.go
```diff
@@ -18,505 +18,339 @@ package hyperparametertuningjob
...
 		}
 
-		return r.removeFinalizerAndUpdate(ctx)
+	case sagemaker.HyperParameterTuningJobStatusStopping:
+		break
+
 	default:
-		log.Info("Job is in unknown status")
-		return NoRequeue()
+		unknownStateError := errors.New(fmt.Sprintf("Unknown Tuning Job Status: %s", ctx.TuningJobDescription.HyperParameterTuningJobStatus))
```
cadedaniel: We should use `fmt.Errorf(...)` instead of `errors.New(fmt.Sprintf(...))`. I know you were just copying from old TrainingJob, this was a thing that I learned between writing TJ and Hosting.

controllers/hyperparametertuningjob/hyperparametertuningjob_controller_test.go
```diff
@@ -1701,34 +661,153 @@ func createHpoJob(withFinalizer bool, k8sName, k8sNamespace, sageMakerName, regi
...
+
+	calledJob := spawner.spawnMissingTrainingJobsCalls.Front().Value.(hpojobv1.HyperparameterTuningJob)
+	Expect(*calledJob.Spec.Region).To(Equal(*job.Spec.Region))
+	Expect(calledJob.ObjectMeta.GetName()).To(Equal(job.ObjectMeta.GetName()))
+	Expect(calledJob.ObjectMeta.GetNamespace()).To(Equal(job.ObjectMeta.GetNamespace()))
+}
+
+// Helper function to verify that the controller attempted to spawn the missing child training jobs.
+func ExpectSpawnMissingTrainingJobs(spawner mockTrackingHPOTrainingJobSpawner, calls int, job hpojobv1.HyperparameterTuningJob) {
```
cadedaniel: The naming choices here make usage of this function confusing. These questions came to my mind when encountering this in one of the tests: What does `1` mean? Does this assert all calls for the function, or just one? Is it testing the call parameters, or just that it was called?

We should either have it specifically look at the first call, or take a call index (_a la_ [ExpectNthSubreconcilerCallToKeepUnusedResources](https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/controllers/hosting/hostingdeployment_controller_test.go#L1036)). We should also pick a function name that better fits what it does, maybe something like `ExpectNthSpawnMissingTrainingJobsCallToHaveHPOJob`? I'm open to better names as long as they convey the purpose reasonably well.

Same goes for `ExpectDeletedSpawnedTrainingJobs`.

controllers/sdkutil/spec_comparison_test.go
```diff
@@ -258,175 +258,176 @@ var _ = Describe("HyperParameterTuningJobMatchesDescription", func() {
...
-			name2: value2,
-		}
-
-		comparison := TrainingJobSpecMatchesDescription(description, spec)
-
-		Expect(comparison.Equal).To(Equal(true))
-		Expect(comparison.Differences).To(Equal(""))
-	})
-})
+// TODO: Ensure there are no loose ends when deleting the TrainingJobSpecMatchesDescription method.
```
cadedaniel: 🎉🎉🎉

PR #87, opened 2020-01-30
Modified canary common.sh source path

RedbackThomson:
Considering the current directory context is different between the integration test and the canary test, the source was only correct for the integration tests. I have modified the path of the `source common.sh` to relatively import the file, since `run_canarytest.sh` is activated in the same directory.

I have tested this change by running the container on my local machine and ensuring it is able to run the `delete_all_resources` method.

### DIFF
--- a/tests/codebuild/run_canarytest.sh
+++ b/tests/codebuild/run_canarytest.sh
@@ -1,18 +1,5 @@
 #!/bin/bash
 
-source ./common.sh
-
-# TODOs
-# 1. Add validation for each steps and abort the test if steps fails
-# Build environment `Docker image` has all prerequisite setup and credentials are being passed using AWS system manager
-
-CLUSTER_REGION=${CLUSTER_REGION:-us-east-1}
-CLUSTER_VERSION=${CLUSTER_VERSION:-1.13}
-
-# Define the list of optional subnets for the EKS test cluster
-CLUSTER_PUBLIC_SUBNETS=${CLUSTER_PUBLIC_SUBNETS:-}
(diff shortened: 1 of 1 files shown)


tests/codebuild/run_canarytest.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-source tests/codebuild/common.sh
+source ./common.sh
```
cadedaniel: You mentioned something about `set -e` flag in chat.. would that have helped us catch this // should we set it?

RedbackThomson: It's not shown in this pull request preview, but we `set -e` on line 46 after we create the cleanup trap. I will move this source down below that to catch this in the future.

cadedaniel: Hmm I see how that would make "imports" hairy.

tests/codebuild/run_canarytest.sh
```diff
@@ -1,18 +1,5 @@
...
-
-CLUSTER_REGION=${CLUSTER_REGION:-us-east-1}
-CLUSTER_VERSION=${CLUSTER_VERSION:-1.13}
-
-# Define the list of optional subnets for the EKS test cluster
-CLUSTER_PUBLIC_SUBNETS=${CLUSTER_PUBLIC_SUBNETS:-}
-CLUSTER_PRIVATE_SUBNETS=${CLUSTER_PRIVATE_SUBNETS:-}
-
 # Verbose trace of commands, helpful since test iteration takes a long time.
 set -x
```
cadedaniel: You usually want `set -x` after `source`, otherwise it produces a lot of output from sourcing the file.

PR #88, opened 2020-02-03
Update Training job Status and Spec

RedbackThomson:
In the `smlogs` code, we reference the training job name from the spec. The training job name was not being updated in the spec or in the status. This is not caught in unit tests because it tests interoperability between smlogs and the controller.

### DIFF
--- a/controllers/trainingjob/trainingjob_controller.go
+++ b/controllers/trainingjob/trainingjob_controller.go
@@ -245,12 +245,6 @@ func (r *Reconciler) initializeContext(ctx *reconcileRequestContext) error {
 func (r *Reconciler) initializeTrainingJobName(ctx *reconcileRequestContext) error {
 	ctx.TrainingJob.Spec.TrainingJobName = &ctx.TrainingJobName
 
-	ctx.TrainingJob.Status.SageMakerTrainingJobName = ctx.TrainingJobName
-	//TODO: Convert it to tinyurl or even better can we expose CW url via API server proxy UI?
-	ctx.TrainingJob.Status.CloudWatchLogUrl = "https://" + *ctx.TrainingJob.Spec.Region + ".console.aws.amazon.com/cloudwatch/home?region=" +
-		*ctx.TrainingJob.Spec.Region + "#logStream:group=/aws/sagemaker/TrainingJobs;prefix=" +
-		ctx.TrainingJobName + ";streamFilter=typeLogStreamPrefix"
-
 	if err := r.Status().Update(ctx, ctx.TrainingJob); err != nil {
 		ctx.Log.Error(err, "Error while updating training job name")
 		return err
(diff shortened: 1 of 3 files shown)


controllers/trainingjob/trainingjob_controller.go
```diff
@@ -226,6 +226,7 @@ func (r *Reconciler) initializeContext(ctx *reconcileRequestContext) error {
 		ctx.TrainingJobName = *ctx.TrainingJob.Spec.TrainingJobName
 	} else {
 		ctx.TrainingJobName = controllers.GetGeneratedJobName(ctx.TrainingJob.ObjectMeta.GetUID(), ctx.TrainingJob.ObjectMeta.GetName(), MaxTrainingJobNameLength)
+		ctx.TrainingJob.Spec.TrainingJobName = &ctx.TrainingJobName
```
cadedaniel: We should call `r.client.Update` here, to guarantee that the spec name is persisted to etcd.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -308,6 +309,9 @@ func (r *Reconciler) updateStatusWithAdditional(ctx reconcileRequestContext, tra
 	ctx.Log.Info("updateStatusWithAdditional", "trainingJobPrimaryStatus", trainingJobPrimaryStatus, "trainingJobSecondaryStatus", trainingJobSecondaryStatus, "additional", additional)
 
 	jobStatus := &ctx.TrainingJob.Status
+
+	jobStatus.SageMakerTrainingJobName = ctx.TrainingJobName
```
cadedaniel: This feels unnecessarily complicated. We can just do the `Update` where we change the name (once per job).

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -227,6 +234,14 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
 					createdRequest := req.(*sagemaker.CreateTrainingJobInput)
 					Expect(*createdRequest.TrainingJobName).To(Equal("training-job-name"))
 				})
+
+				It("Does not modify the job name in the spec", func() {
+					ExpectTrainingJobNameInSpec("training-job-name", trainingJob)
```
cadedaniel: nit: The "training-job-name" should be kept in a constant local to this `Context` block.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -215,6 +214,14 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
 				ExpectStatusToBe(trainingJob, string(sagemaker.TrainingJobStatusInProgress), string(sagemaker.SecondaryStatusStarting))
 			})
 
+			It("Adds the training job name to the spec", func() {
+				ExpectTrainingJobNameInSpec(controllers.GetGeneratedJobName(trainingJob.ObjectMeta.GetUID(), trainingJob.ObjectMeta.GetName(), MaxTrainingJobNameLength), trainingJob)
+			})
+
+			It("Adds the training job name to the status", func() {
```
cadedaniel: Can add test using [`FailToUpdateK8sClient`](https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/controllers/controllertest/k8s_mocks.go#L92), which verifies that `reconcileResult` is `requeueImmediately` when the update fails (it is simple, just needs a Context/BeforeEach which sets `kubernetesClient = FailToUpdateK8sClient{ActualClient: kubernetesClient}`.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -165,6 +165,19 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
 		})
 	})
 
+	Context("K8s client fails to update", func() {
```
cadedaniel: Change to `K8s client fails to update generated spec name` so that debugging is easier when this fails.

PR #89, opened 2020-02-05
Canary fix for delete job status

mbaijal:
Thanks @cadedaniel for the fix.
Locally ran the unit test but could not verify the canary tests themselves. 

### DIFF
--- a/tests/codebuild/common.sh
+++ b/tests/codebuild/common.sh
@@ -19,7 +19,7 @@ function get_sagemaker_model_from_k8s_model()
 #    $2: Instance of CRD
 #    $3: Timeout to complete the test
 #    $4: The status that verifies the job has succeeded.
-function wait_for_crd_status_else_fail()
+function wait_for_crd_status()
 {
   local crd_type="$1"
   local crd_instance="$2"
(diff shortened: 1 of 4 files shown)


tests/codebuild/common.sh
```diff
@@ -33,7 +33,7 @@ function wait_for_crd_status_else_fail()
 
   if [ $? -ne 0 ]; then
     echo "[FAILED] ${crd_type} ${crd_instance} did not change status to ${desired_status} within ${timeout}"
-    exit 1
+    return 1 # TODO need to change all code that used this
```
cadedaniel: Remove comment? I am assuming that you searched for all usage of this function and modified them

tests/codebuild/create_tests.sh
```diff
@@ -107,7 +111,10 @@ function verify_test()
   fi
 
   echo "Waiting for job to complete"
-  wait_for_crd_status_else_fail "$crd_type" "$crd_instance" "$timeout" "$desired_status"
+  if ! wait_for_crd_status_else_fail "$crd_type" "$crd_instance" "$timeout" "$desired_status"; then
+      echo "[FAILED] Waiting for status Completed failed"
```
cadedaniel: Can we remove the log from the function so that this message is not duplicated? It will make our usage of `wait_for_crd_status_else_fail` more consistent, since printing the failure message is owned by the caller.

tests/codebuild/common.sh
```diff
@@ -33,7 +33,7 @@ function wait_for_crd_status_else_fail()
```
cadedaniel: Can we rename the function, now that it doesn't `exit`? `wait_for_crd_status` I think is apt.

tests/codebuild/common.sh
mbaijal: Yes I did.

PR #90, opened 2020-02-06
Deduplicate the logic for resource name generation

mbaijal:
Deduplicate logic that calculates generated names, truncating them for certain lengths. 

Testing:
All Unit tests have passed. 

### DIFF
--- a/controllers/aws_config_loader.go
+++ b/controllers/aws_config_loader.go
@@ -28,29 +28,29 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Simple struct to facilitate loading AWS config with region- and endpoint-overrides.
+// AwsConfigLoader is a simple struct to facilitate loading AWS config with region- and endpoint-overrides.
 // This uses venv.Env for mocking in tests.
 type AwsConfigLoader struct {
 	Env venv.Env
 }
 
-// Create an AwsConfigLoader with the default OS environment.
+// NewAwsConfigLoader creates an AwsConfigLoader with the default OS environment.
 func NewAwsConfigLoader() AwsConfigLoader {
 	return NewAwsConfigLoaderForEnv(venv.OS())
 }
 
+// NewAwsConfigLoaderForEnv returns a AwsConfigLoader for the specified environment.
 func NewAwsConfigLoaderForEnv(env venv.Env) AwsConfigLoader {
 	return AwsConfigLoader{
 		Env: env,
 	}
 }
 
-/*
-   As of 11/16/2019 aws-sdk-go-v2 does not supports AWS_WEB_IDENTITY_TOKEN_FILE
-   based credentials. This is a work around to retrieve the credentials from v1
-   and pass to v2.
-   TODO: Remove this function once aws-sdk-go-v2 is fixed.
-*/
+// InstallCredsUsingSDKV1 is required because as of 11/16/2019 aws-sdk-go-v2 does not
+// supports AWS_WEB_IDENTITY_TOKEN_FILE based credentials. This is a work around to
+// retrieve the credentials from v1 and pass to v2.
+// TODO: Remove this function once aws-sdk-go-v2 is fixed.
+// TODO: Change var awsAccessKeyId to awsAccessKeyID
 func (l *AwsConfigLoader) InstallCredsUsingSDKV1(config *aws.Config, regionOverride string) {
 	var err error
 	awsAccessKeyId := l.Env.Getenv("AWS_ACCESS_KEY_ID")
@@ -90,7 +90,7 @@ func (l *AwsConfigLoader) InstallCredsUsingSDKV1(config *aws.Config, regionOverr
 	}
 }
 
-// Load default AWS config and apply overrides, like setting the region and using a custom SageMaker endpoint.
+// LoadAwsConfigWithOverrides loads default AWS config and apply overrides, like setting the region and using a custom SageMaker endpoint.
 // If specified, jobSpecificEndpointOverride always overrides the endpoint. Otherwise, the environment
 // variable specified by DefaultSageMakerEndpointEnvKey overrides the endpoint if it is set.
 func (l AwsConfigLoader) LoadAwsConfigWithOverrides(regionOverride string, jobSpecificEndpointOverride *string) (aws.Config, error) {
--- a/controllers/batchtransformjob/batchtransformjob_controller.go
+++ b/controllers/batchtransformjob/batchtransformjob_controller.go
@@ -39,7 +39,12 @@ import (
 
 // The HTTP message returned with HTTP code 400 if a DescribeTransformJob request
 // cannot find the given transform job.
-const transformResourceNotFoundApiCode string = "ValidationException"
+const (
+	transformResourceNotFoundApiCode string = "ValidationException"
+
+	// Defines the maximum number of characters in a SageMaker Endpoint Config Resource name
+	MaxBatchTransformNameLength = 63
+)
 
 // BatchTransformJobReconciler reconciles a BatchTransformJob object
 type BatchTransformJobReconciler struct {
@@ -262,7 +267,7 @@ func (r *BatchTransformJobReconciler) addFinalizerAndRequeue(ctx reconcileReques
 }
 
 func (r *BatchTransformJobReconciler) getTransformJobName(state batchtransformjobv1.BatchTransformJob) string {
-	return GetGeneratedJobName(state.ObjectMeta.GetUID(), state.ObjectMeta.GetName(), 63)
+	return GetGeneratedJobName(state.ObjectMeta.GetUID(), state.ObjectMeta.GetName(), MaxBatchTransformNameLength)
 }
 
 func (r *BatchTransformJobReconciler) reconcileSpecWithDescription(ctx reconcileRequestContext) (ctrl.Result, error) {
--- a/controllers/batchtransformjob/batchtransformjob_controller.go
+++ b/controllers/batchtransformjob/batchtransformjob_controller.go
@@ -42,7 +42,7 @@ import (
 const (
 	transformResourceNotFoundApiCode string = "ValidationException"
 
-	// Defines the maximum number of characters in a SageMaker Endpoint Config Resource name
+	// Defines the maximum number of characters in a SageMaker Batch Transform Resource name
 	MaxBatchTransformNameLength = 63
 )
 
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -31,32 +31,33 @@ import (
 )
 
 const (
-	// The name of the finalizer that is installed onto managed etcd resources.
+	// SageMakerResourceFinalizerName is the name of the finalizer that is installed onto managed etcd resources.
 	SageMakerResourceFinalizerName = "sagemaker-operator-finalizer"
 
-	// The environment variable key of the default SageMaker endpoint to use.
+	// DefaultSageMakerEndpointEnvKey is the environment variable key of the default SageMaker endpoint to use.
 	// If this is set to anything other than the empty string, the controllers
 	// will use its value as the endpoint with which to communicate with SageMaker.
 	DefaultSageMakerEndpointEnvKey = "AWS_DEFAULT_SAGEMAKER_ENDPOINT"
 
-	// Status stored on first touch of a job by a controller.
+	// InitializingJobStatus is the status stored on first touch of a job by a controller.
 	// This status is kept until the job has a real status.
 	InitializingJobStatus = "SynchronizingK8sJobWithSageMaker"
 )
 
-// The error message to use when a spec differs from its SageMaker description.
+// CreateSpecDiffersFromDescriptionErrorMessage returns the error message to use when a spec differs from its SageMaker description.
 func CreateSpecDiffersFromDescriptionErrorMessage(job interface{}, status, differences string) string {
 	const specDiffersFromDescriptionErrorMessage = "Updates to %s are not supported; the resource no longer matches the SageMaker resource. The job will be marked as \"%s\" and no modifications to the existing SageMaker job will be made until the update is reverted. The differences that prevent tracking of the SageMaker resource are:\n%s"
 	return fmt.Sprintf(specDiffersFromDescriptionErrorMessage, reflect.TypeOf(job).String(), status, differences)
 }
 
-// This string will be added to user agent. This helps to distinguish
+// SagemakerOnKubernetesUserAgentAddition is the string that will be added to user agent. This helps to distinguish
 // sagemaker job which has been created from kubernetes.
 const SagemakerOnKubernetesUserAgentAddition = "sagemaker-on-kubernetes"
 
-// Helper type that describes the action to perform on a resource.
+// ReconcileAction is a helper type that describes the action to perform on a resource.
 type ReconcileAction string
 
+// Constants for the various reconcile actions.
 const (
 	NeedsCreate ReconcileAction = "NeedsCreate"
 	NeedsDelete ReconcileAction = "NeedsDelete"
@@ -64,23 +65,23 @@ const (
 	NeedsUpdate ReconcileAction = "NeedsUpdate"
 )
 
-// Type for function that returns a SageMaker client. Used for mocking.
+// SageMakerClientProvider is a Type for function that returns a SageMaker client. Used for mocking.
 type SageMakerClientProvider func(aws.Config) sagemakeriface.ClientAPI
 
+// RequeueIfError requeues if an error is found.
 func RequeueIfError(err error) (ctrl.Result, error) {
 	return ctrl.Result{}, err
 }
 
-// Helper function which requeues immediately if the object generation has not changed.
+// RequeueImmediatelyUnlessGenerationChanged is a helper function which requeues immediately if the object generation has not changed.
 // Otherwise, since the generation change will trigger an immediate update anyways, this
 // will not requeue.
 // This prevents some cases where two reconciliation loops will occur.
 func RequeueImmediatelyUnlessGenerationChanged(prevGeneration, curGeneration int64) (ctrl.Result, error) {
 	if prevGeneration == curGeneration {
 		return RequeueImmediately()
-	} else {
-		return NoRequeue()
 	}
+	return NoRequeue()
 }
 
 // Note: In unit test we use Result from controller-runtime package from https://github.com/kubernetes-sigs/controller-runtime/blob/2027a413747f2a8cada813dd98b3b1473c253913/pkg/reconcile/reconcile.go#L26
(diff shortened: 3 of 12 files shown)


controllers/common_test.go
```diff
@@ -148,4 +152,41 @@ var _ = Describe("GetGeneratedJobName", func() {
...
+			Expect(len(generatedJobNameWithPostfix)).To(BeNumerically("<=", maxNameLen))
+		})
+	})
+
+	When("An additional postfix is required and maxNameLen is not large enough for objectMetaName", func() {
+		BeforeEach(func() {
+			maxNameLen = 64
+			objectMetaName = strings.Repeat("A", 70)
+			generatedPostfix = "generated.postfix"
+			fmt.Printf("Hey Meghna, the generated name is : %s", generatedJobNameWithPostfix)
```
RedbackThomson: Another debugging mark?

controllers/common_test.go
```diff
@@ -148,4 +152,41 @@ var _ = Describe("GetGeneratedJobName", func() {
...
+		})
+
+		It("Concatenates the name, shortened uid, generatedPostfix", func() {
```
RedbackThomson: nit: Update since this doesn't check it concatenates the name

controllers/common.go
```diff
@@ -152,10 +152,11 @@ func Now() *metav1.Time {
 // * [CreateTrainingJob#TrainingJobName](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateTrainingJob.html#SageMaker-CreateTrainingJob-request-TrainingJobName)
 // * [CreateTransformJob#TransformJobName](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateTransformJob.html#SageMaker-CreateTransformJob-request-TransformJobName)
 // * [CreateHyperParameterTuningJob#HyperParameterTuningJobName](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateHyperParameterTuningJob.html#SageMaker-CreateHyperParameterTuningJob-request-HyperParameterTuningJobName)
-func GetGeneratedJobName(objectMetaUID types.UID, objectMetaName string, maxNameLen int) string {
-	requiredPostfix := strings.Replace(string(objectMetaUID), "-", "", -1)
-
+func GetGeneratedResourceName(objectMetaUID types.UID, objectMetaName string, maxNameLen int, generatedPostfix ...string) string {
+	uid := strings.Replace(string(objectMetaUID), "-", "", -1)
+	requiredPostfix := strings.Join(generatedPostfix, "") + uid
```
RedbackThomson: Could we instead join the requiredPostfix with a delimiter before the uid so that when we use the generatedPostfix we don't need to remember the hyphen?

For example:
`requiredPostfix := strings.Join(generatedPostfix, "") + delimiter + uid`

controllers/hosting/hostingdeployment_controller.go
```diff
@@ -468,22 +469,28 @@ func convertTimeOrDefault(time *time.Time, defaultValue *metav1.Time) *metav1.Ti
...
+
+// Get the Kubernetes name of the Model/EndpointConfig. This must be idempotent so that future reconciler invocations
+// are able to find the object.
+// Kubernetes resources can have names up to 253 characters long.
+// The characters allowed in names are: digits (0-9), lower case letters (a-z), -, and .
+func GetKubernetesNamespacedName(objectName string, hostingDeployment hostingv1.HostingDeployment) types.NamespacedName {
+	k8sMaxLen := 253
+	uid := hostingDeployment.ObjectMeta.GetUID()
+	generation := strconv.FormatInt(hostingDeployment.ObjectMeta.GetGeneration(), 10)
+	generatedPostfix := generation + "-"
```
RedbackThomson: See comment on line 157 on `common.go`

controllers/hosting/hostingdeployment_controller_test.go
```diff
@@ -40,6 +40,7 @@ import (
 	commonv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/common"
 	endpointconfigv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/endpointconfig"
 	hostingv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/hostingdeployment"
+	. "github.com/aws/amazon-sagemaker-operator-for-k8s/controllers"
```
RedbackThomson: Please avoid using dot imports - they give warnings in golint. I've been trying to remove our use of these in the refactor work.

controllers/endpointconfig/endpointconfig_controller.go
```diff
@@ -150,7 +149,7 @@ func (r *EndpointConfigReconciler) reconcileEndpointConfig(ctx reconcileRequestC
 	}
 
 	// Get SageMaker endpointConfig description.
-	if ctx.EndpointConfigDescription, err = ctx.SageMakerClient.DescribeEndpointConfig(ctx, generateEndpointConfigName(ctx.EndpointConfig)); err != nil {
+	if ctx.EndpointConfigDescription, err = ctx.SageMakerClient.DescribeEndpointConfig(ctx, GetGeneratedResourceName(ctx.EndpointConfig.ObjectMeta.GetUID(), ctx.EndpointConfig.ObjectMeta.GetName(), 63)); err != nil {
```
RedbackThomson: Would you be able to create constants for holding the name lengths for each resource? I've included this in the TrainingJob refactor, as an example.

controllers/model/model_controller.go
```diff
@@ -146,7 +145,7 @@ func (r *ModelReconciler) reconcileModel(ctx reconcileRequestContext) error {
 	}
 
 	// Get SageMaker model description.
-	if ctx.ModelDescription, err = ctx.SageMakerClient.DescribeModel(ctx, generateModelName(ctx.Model)); err != nil {
+	if ctx.ModelDescription, err = ctx.SageMakerClient.DescribeModel(ctx, GetGeneratedResourceName(ctx.Model.ObjectMeta.GetUID(), ctx.Model.ObjectMeta.GetName(), 63)); err != nil {
```
cadedaniel: Can we deduplicate the 3 identical calls to GetGeneratedResourceName?

controllers/common_test.go
```diff
@@ -148,4 +152,41 @@ var _ = Describe("GetGeneratedJobName", func() {
 			Expect(uidWithoutHyphens).To(ContainSubstring(generatedJobName))
 		})
 	})
+
+	When("An additional postfix is required and maxNameLen is sufficiently large", func() {
```
cadedaniel: Reminder for me to look at tests in-depth on second pass.

controllers/common.go
mbaijal: > Could we instead join the requiredPostfix with a delimiter before the uid so that when we use the generatedPostfix we don't need to remember the hyphen?
> 
> For example:
> `requiredPostfix := strings.Join(generatedPostfix, "") + delimiter + uid`

I did not do this since this would require a conditional check to add the delimiter only when the generatedPostfix does exist.

mbaijal: > We need to be careful of changing the format of these names. Because our installers set the image pull policy to `Always`, we risk breaking a customer setup when we change them. Let's avoid changes here unless we want to have a live-migration code.

The formats of the names are not changing here. The generated names would be the exact same as before. Can you elaborate?

controllers/hosting/hostingdeployment_controller.go
mbaijal: I did this intentionally, check response earlier.

controllers/hosting/hostingdeployment_controller.go
```diff
@@ -468,22 +469,28 @@ func convertTimeOrDefault(time *time.Time, defaultValue *metav1.Time) *metav1.Ti
...
 		// Ignore status-only and metadata-only updates
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// Get the Kubernetes name of the Model/EndpointConfig. This must be idempotent so that future reconciler invocations
+// are able to find the object.
+// Kubernetes resources can have names up to 253 characters long.
+// The characters allowed in names are: digits (0-9), lower case letters (a-z), -, and .
+func GetKubernetesNamespacedName(objectName string, hostingDeployment hostingv1.HostingDeployment) types.NamespacedName {
```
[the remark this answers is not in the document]
mbaijal: I originally added it to the common.go file but leads to a cyclic import for hostingDeployments and is not possible to do. This seemed like a better (logically common) place for the function than putting it in either model of endpoint config and then calling it in the other. Open to moving it though.

controllers/hosting/hostingdeployment_controller.go
```diff
@@ -386,7 +387,7 @@ func (r *HostingDeploymentReconciler) createCreateEndpointInput(ctx reconcileReq
 		return nil, err
 	}
 
-	endpointName := GetSageMakerEndpointName(*ctx.Deployment)
```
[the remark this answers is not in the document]
mbaijal: @mbaijal This is pending

controllers/batchtransformjob/batchtransformjob_controller.go
```diff
@@ -39,7 +39,12 @@ import (
 
 // The HTTP message returned with HTTP code 400 if a DescribeTransformJob request
 // cannot find the given transform job.
-const transformResourceNotFoundApiCode string = "ValidationException"
+const (
+	transformResourceNotFoundApiCode string = "ValidationException"
+
+	// Defines the maximum number of characters in a SageMaker Endpoint Config Resource name
```
mbaijal: Fix comment

controllers/common_test.go
```diff
@@ -36,15 +36,25 @@ var _ = Describe("GetGeneratedJobName", func() {
...
 	})
 
 	JustBeforeEach(func() {
 		uidWithoutHyphens = strings.Replace(string(uid), "-", "", -1)
-		generatedJobName = GetGeneratedJobName(uid, objectMetaName, maxNameLen)
+
+		if shouldJobNameHavePostfix {
+			generatedJobName = GetGeneratedResourceName(generatedPostfix+"-"+uidWithoutHyphens, objectMetaName, maxNameLen)
+		} else {
+			generatedJobName = GetGeneratedJobName(uid, objectMetaName, maxNameLen)
```
cadedaniel: Why are we testing another function in `GetGeneratedResourceName` tests? This couples `GetGeneratedResourceName` with `GetGeneratedJobName`, making code harder to refactor and tests more difficult to comprehend.

I think since `GetGeneratedJobName` merely calls `GetGeneratedResourceName`, we only need a test on the calling logic for `GetGeneratedJobName`. Can we break out a new `Describe` case for `GetGeneratedJobName` which only tests 1) that the UID does not have hyphens and 2) that the output is equivalent to `GetGeneratedResourceName` output for some basic case?

controllers/common_test.go
```diff
@@ -148,4 +158,42 @@ var _ = Describe("GetGeneratedJobName", func() {
 			Expect(uidWithoutHyphens).To(ContainSubstring(generatedJobName))
 		})
 	})
+
+	When("An additional postfix is required and maxNameLen is sufficiently large", func() {
+		BeforeEach(func() {
+			maxNameLen = 253
+			objectMetaName = "object.meta.name"
+			generatedPostfix = "generated.postfix"
```
cadedaniel: This usage is confusing to me. Who generates this postfix? What is it postfixed to?

cadedaniel: I think these tests would be much more clear if they used `required` and `optional` strings, following the parameters of `GetGeneratedResourceName`.

controllers/common_test.go
```diff
@@ -148,4 +158,42 @@ var _ = Describe("GetGeneratedJobName", func() {
...
+			shouldJobNameHavePostfix = true
+
+		})
+
+		It("Concatenates the name, shortened uid, generatedPostfix", func() {
+			Expect(generatedJobName).To(ContainSubstring(objectMetaName))
+			Expect(generatedJobName).To(ContainSubstring(uidWithoutHyphens))
+			Expect(generatedJobName).To(ContainSubstring(generatedPostfix))
```
cadedaniel: Why are we differentiating between different substrings of the `required` parameter? Shouldn't the function's behavior be the same for the entire `required` parameter?

controllers/common.go
cadedaniel: > > We need to be careful of changing the format of these names. Because our installers set the image pull policy to Always, we risk breaking a customer setup when we change them. Let's avoid changes here unless we want to have a live-migration code.
> 
> The formats of the names are not changing here. The generated names would be the exact same as before. Can you elaborate?

I am saying that although Nick's suggestion is a good one, we shouldn't break these names until v2 unless we really need to.

controllers/common_test.go
RedbackThomson: Could we also verify the correct ordering for all the values and that they are delimited in the correct places?

controllers/common_test.go
```diff
@@ -148,4 +158,42 @@ var _ = Describe("GetGeneratedJobName", func() {
...
+
+		It("Concatenates the shortened uid, generatedPostfix", func() {
+			Expect(generatedJobName).To(ContainSubstring(uidWithoutHyphens))
+			Expect(generatedJobName).To(ContainSubstring(generatedPostfix))
+		})
+
+		It("Length does not exceed maxNameLen", func() {
+			Expect(len(generatedJobName)).To(BeNumerically("<=", maxNameLen))
+		})
+	})
```
RedbackThomson: Is there a test for substrings of the required part without a required postfix? Just making sure we hit each condition

controllers/common.go
RedbackThomson: Any changes to these formats would necessitate a major version upgrade.

controllers/common_test.go
mbaijal: did not update the tests very well to match the changed methods, fixing

controllers/common_test.go
mbaijal: I think its still a good idea to have required = postfix+uid in order to check that if maxlen<required, its the uid part that is returned.

controllers/common_test.go
mbaijal: without a required postfix - not sure what you mean. 
There is a test to ensure that if required is larger than the maxlen, uid is prioritizied.

mbaijal:
> Would you mind also checking to see if any of these changes introduced further linting errors? It's been a goal of mine for every PR to reduce the linting error count. I know there are still a lot, so it might be hard to see which of those have been introduced by your changes.

I have fixed all linting errors in the common.go file and a few in aws_config_loader.go (except one for which I added a TODO). The rest of the project looks clean!

controllers/aws_config_loader.go
```diff
@@ -28,29 +28,29 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Simple struct to facilitate loading AWS config with region- and endpoint-overrides.
+// AwsConfigLoader is a simple struct to facilitate loading AWS config with region- and endpoint-overrides.
```
cadedaniel: Thanks again for taking care of these additional lint fixes.

controllers/model/model_controller.go
```diff
@@ -344,3 +334,8 @@ func (r *ModelReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// Given a Model object, generate a Sagemaker name of valid length
```
cadedaniel: nit: `s/Sagemaker/SageMaker`.

controllers/common_test.go
```diff
@@ -133,19 +133,64 @@ var _ = Describe("GetGeneratedJobName", func() {
...
+	)
+
+	BeforeEach(func() {
+		uid = types.UID(uuid.New().String())
+	})
+
+	JustBeforeEach(func() {
+		uidWithoutHyphens = strings.Replace(string(uid), "-", "", -1)
+		generatedJobName = GetGeneratedJobName(uid, objectMetaName, maxNameLen)
+		generatedResourceName = GetGeneratedJobName(uid, objectMetaName, maxNameLen)
```
cadedaniel: These variables are assigned the same value? I think you mean to call `GetGeneratedResourceName` with `uidWithoutHyphens`? This makes the test on 192 a very happy test that won't ever fail :)

If you are only using `generatedResourceName` in one test, it might make more sense to create it in that test alone.

controllers/common_test.go
```diff
@@ -133,19 +133,64 @@ var _ = Describe("GetGeneratedJobName", func() {
...
+
+		It("Equals the truncated required string", func() {
+			Expect(generatedResourceName).To(Equal(required[:maxNameLen]))
+		})
+
+		It("Length does not exceed maxNameLen", func() {
+			Expect(len(generatedResourceName)).To(BeNumerically("<=", maxNameLen))
+		})
+
+		It("Is contained in the full hyphen-less UID", func() {
```
cadedaniel: nit: update to say `Is contained in the full required string`.

controllers/common_test.go
```diff
@@ -44,63 +43,64 @@ var _ = Describe("GetGeneratedJobName", func() {
 
 	JustBeforeEach(func() {
 		uidWithoutHyphens = strings.Replace(string(uid), "-", "", -1)
```
cadedaniel: nit: It is weird that the `uidWithoutHyphens` is still here, since you refactored out the logic that removes hyphens. It will make test easier to understand if you remove it since readers will not have to understand this element -- the function `GetGeneratedResourceName` should work for arbitrary string input.

controllers/common_test.go
cadedaniel: We can do the additional tests Nick suggests, but given that we don't have any requirement on the ordering I think we are good as is.

controllers/common_test.go
mbaijal: oh yes, good catch. Fixed. 

It is a little tricky to move the creation of `generatedResourceName` since the `uid` gets generated a bunch of times for each spec. 
I tried to create the `generatedResourceName` within the `Before Each` block of the test itself (inside the `When`). 
This causes the `generatedJobName` and `GetGeneratedResourceName` to be called with different `uid` and `uidWithoutHyphens`. An example of the failure log below: 

```
•••••••••••••generatedJobName is object.meta.name-bced7caf409947faa7d2161963a1bff7
generatedResourceName is : object.meta.name-
•generatedJobName is object.meta.name-3c0b1cadf1e6465d9043f791d56f8d46
generatedResourceName is : object.meta.name-bced7caf409947faa7d2161963a1bff7
•generatedJobName is object.meta.name-2e70d1f94c12493f8797310821130c58
generatedResourceName is : object.meta.name-3c0b1cadf1e6465d9043f791d56f8d46
---------------------
    Expected
        <string>: object.meta.name-2e70d1f94c12493f8797310821130c58
    to be equivalent to
        <string>: object.meta.name-3c0b1cadf1e6465d9043f791d56f8d46
```

controllers/common_test.go
mbaijal: It does, but the idea was to check if the length of required is more than maxLen, the second half of required which contains the uid gets picked. I changed this to not use a generated UID but a hardcoded one. Let me know if you prefer to remove this check altogether.

controllers/common_test.go
cadedaniel: Hmm.. I don't know how this issue was caused. It seems that the two `BeforeEach`s are called in the wrong order, which means that the UID from the last test and this test are used.

I wouldn't put it in a `BeforeEach` (which runs before a set of tests), and simply place it in the single test where it is used, i.e.:

```go
It("returns the same name as GetGeneratedResourceName with uidWithoutHyphens as the required string", func() {
        generatedResourceName := GetGeneratedResourceName(uidWithoutHyphens, objectMetaName, maxNameLen)
	Expect(generatedJobName).To(BeEquivalentTo(generatedResourceName))
})
```

This is pretty minor, so fine leaving as is.

controllers/common_test.go
cadedaniel: > It does, but the idea was to check if the length of required is more than maxLen, the second half of required which contains the uid gets picked

This test is fine but somewhat imprecise -- if `GetGeneratedResourceName` has room only for `required` (no `optional`), why would our tests only check that the last half of `required` is present?

To make the test better match the documented behavior of the function, we should change:
```go
It("Contains the full UID", func() {
    Expect(generatedResourceName).To(ContainSubstring(requiredPostfix))
})
```
to:
```go
It("Contains the full required string", func() {
    Expect(generatedResourceName).To(ContainSubstring(required))
})
```

The impact of this change pretty small so happy to approve either way.

cadedaniel:
Can we make sure to run integration tests or smoke tests on this? I think if you do a manual HPO job and manual HostingDeployment it will be a good enough test since HPO spawns TrainingJobs and HostingDeployment is the only one that uses the Kubernetes names.

mbaijal:
> Can we make sure to run integration tests or smoke tests on this? I think if you do a manual HPO job and manual HostingDeployment it will be a good enough test since HPO spawns TrainingJobs and HostingDeployment is the only one that uses the Kubernetes names.

I did a manual [1]hostingDeployment, [2]trainingJob, [3]BatchTransform and [4]HPO. I verified the endpoint, endpointconfig, model names for [1]. [2] and [3] did not go to completion due to incorrect specs but were successfully launched.  For [4] training jobs were successfully spawned and are in progress.

PR #96, opened 2020-03-04
Remove container `latest` tag

RedbackThomson:
Removes any reference to the `latest` tag from our installation methods, and from the CD release scripts. Currently, if we were to push a major version update to the master branch, all valid installations would break due to backwards compatibility breaking from v1 to v2. This fix mitigates that by removing access to the `latest` tag, and recommending the latest major version tag as the default installation method.

Fixes # N/A

### All Submissions:

### DIFF
--- a/codebuild/scripts/package_operators.sh
+++ b/codebuild/scripts/package_operators.sh
@@ -34,23 +34,18 @@ function deploy_from_alpha()
   # Login to the prod repository
   $(aws ecr get-login --no-include-email --region ${account_region} --registry-ids ${account_id})
 
-  # Clone the controller image to the repo and set as latest
+  # Clone the controller image to the repo
   docker tag $alpha_ecr_image:$CODEBUILD_RESOLVED_SOURCE_VERSION ${dest_ecr_image}:$CODEBUILD_RESOLVED_SOURCE_VERSION
-  docker tag $alpha_ecr_image:$CODEBUILD_RESOLVED_SOURCE_VERSION ${dest_ecr_image}:latest
 
   # Push to the prod region
   docker push ${dest_ecr_image}:$CODEBUILD_RESOLVED_SOURCE_VERSION
-  docker push ${dest_ecr_image}:latest
 
   printf -v bucket_name $RELEASE_BUCKET_NAME_FMT $RELEASE_TARBALL_BUCKET_PREFIX $account_region
   printf -v binary_prefix $RELEASE_BINARY_PREFIX_FMT $bucket_name
 
-  # Copy across the binaries and set as latest
+  # Copy across the binaries
   aws s3 cp "$ALPHA_LINUX_BINARY_PATH" "$(printf $RELEASE_LINUX_BINARY_PATH_FMT $binary_prefix $CODEBUILD_RESOLVED_SOURCE_VERSION)" $PUBLIC_CP_ARGS
-  aws s3 cp "$ALPHA_LINUX_BINARY_PATH" "$(printf $RELEASE_LINUX_BINARY_PATH_FMT $binary_prefix latest)" $PUBLIC_CP_ARGS
-
   aws s3 cp "$ALPHA_DARWIN_BINARY_PATH" "$(printf $RELEASE_DARWIN_BINARY_PATH_FMT $binary_prefix $CODEBUILD_RESOLVED_SOURCE_VERSION)" $PUBLIC_CP_ARGS
-  aws s3 cp "$ALPHA_DARWIN_BINARY_PATH" "$(printf $RELEASE_DARWIN_BINARY_PATH_FMT $binary_prefix latest)" $PUBLIC_CP_ARGS
 }
 
 # This function builds, packages and deploys a region-specific operator to an ECR repo and output bucket.
--- a/codebuild/scripts/package_operators.sh
+++ b/codebuild/scripts/package_operators.sh
@@ -36,6 +36,10 @@ function deploy_from_alpha()
 
   # Clone the controller image to the repo
   docker tag $alpha_ecr_image:$CODEBUILD_RESOLVED_SOURCE_VERSION ${dest_ecr_image}:$CODEBUILD_RESOLVED_SOURCE_VERSION
+  # TODO: Remove at the time of the next major version release
+  # This moves the `latest` tag along with master branch, which will cause backwards incompatibility for any major
+  # version updates.
+  docker tag $alpha_ecr_image:$CODEBUILD_RESOLVED_SOURCE_VERSION ${dest_ecr_image}:latest
 
   # Push to the prod region
   docker push ${dest_ecr_image}:$CODEBUILD_RESOLVED_SOURCE_VERSION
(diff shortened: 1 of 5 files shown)


config/base/kustomization.yaml
```diff
@@ -10,7 +10,7 @@ patchesStrategicMerge:
 images:
 - name: controller
   newName: 957583890962.dkr.ecr.us-east-1.amazonaws.com/amazon-sagemaker-operator-for-k8s
-  newTag: latest
+  newTag: v1
```
cadedaniel: man.. a v2 update is going to be a lot of build system work.

hack/charts/installer/rolebased/values.yaml
```diff
@@ -2,5 +2,5 @@ roleArn: arn:aws:iam::123456789012:role/DELETE_ME
 
 image:
   repository: 957583890962.dkr.ecr.us-east-1.amazonaws.com/amazon-sagemaker-operator-for-k8s
-  tag: latest
```
cadedaniel: We need to change our current installer [yaml](https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/release/rolebased/installer.yaml#L4014) and [helm chart](https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/hack/charts/installer/rolebased/values.yaml#L5).

codebuild/scripts/package_operators.sh
```diff
@@ -34,23 +34,18 @@ function deploy_from_alpha()
   # Login to the prod repository
   $(aws ecr get-login --no-include-email --region ${account_region} --registry-ids ${account_id})
 
-  # Clone the controller image to the repo and set as latest
+  # Clone the controller image to the repo
   docker tag $alpha_ecr_image:$CODEBUILD_RESOLVED_SOURCE_VERSION ${dest_ecr_image}:$CODEBUILD_RESOLVED_SOURCE_VERSION
-  docker tag $alpha_ecr_image:$CODEBUILD_RESOLVED_SOURCE_VERSION ${dest_ecr_image}:latest
```
cadedaniel: We should keep this `latest` tag until we have `v2`. Otherwise, customers with operators installed before your changes will stop getting minor version updates.

cadedaniel:
Going to also add @gautamkmr to review this since it is an important change.

RedbackThomson:
> What is our test plan for this? I think integration tests + canaries are using pinned tags, so they shouldn't be affected. We should make sure that customers with operators installed at `latest` tag continue to get hotfixes.

Are you asking how we plan to test a v1->v2 migration? Or how we test the changes to the release scripts? Ideally this is where we would have a dev pipeline set up where we can run these scripts in an isolated environment but that is not available yet.

config/base/kustomization.yaml
RedbackThomson: It has all been documented in the MCM.

cadedaniel:
> > What is our test plan for this? I think integration tests + canaries are using pinned tags, so they shouldn't be affected. We should make sure that customers with operators installed at `latest` tag continue to get hotfixes.
> 
> Are you asking how we plan to test a v1->v2 migration? Or how we test the changes to the release scripts? Ideally this is where we would have a dev pipeline set up where we can run these scripts in an isolated environment but that is not available yet.

No, I mean how can we be confident that this won't impact live customers / canaries. I think canaries won't break because they're using tagged ECR images -- they don't depend on this. So this is something specific to the customer installation YAML, not covered by canaries.. we should do a quick smoke test that the release YAML/helm charts work after the change has been deployed.

Or do I understand incorrectly, and canaries do actually test the release YAML we have published?

RedbackThomson:
> No, I mean how can we be confident that this won't impact live customers / canaries. I think canaries won't break because they're using tagged ECR images -- they don't depend on this. So this is something specific to the customer installation YAML, not covered by canaries.. we should do a quick smoke test that the release YAML/helm charts work after the change has been deployed.
> 
> Or do I understand incorrectly, and canaries do actually test the release YAML we have published?

You are correct in that the canaries use an installation tarball which is created using the SHA of the latest release version.

I will quickly set up my cluster to use these tags to ensure they pass all integration tests.

RedbackThomson:
> > No, I mean how can we be confident that this won't impact live customers / canaries. I think canaries won't break because they're using tagged ECR images -- they don't depend on this. So this is something specific to the customer installation YAML, not covered by canaries.. we should do a quick smoke test that the release YAML/helm charts work after the change has been deployed.
> > Or do I understand incorrectly, and canaries do actually test the release YAML we have published?
> 
> You are correct in that the canaries use an installation tarball which is created using the SHA of the latest release version.
> 
> I will quickly set up my cluster to use these tags to ensure they pass all integration tests.

I have just tried both the `release/rolebased` yaml and the new helm charts and both successfully started and tracked an HPO job as in line with `v1.0.1`.

config/base/kustomization.yaml
cadedaniel: Thanks!

codebuild/scripts/package_operators.sh
```diff
@@ -34,23 +34,18 @@ function deploy_from_alpha()
...
   printf -v bucket_name $RELEASE_BUCKET_NAME_FMT $RELEASE_TARBALL_BUCKET_PREFIX $account_region
   printf -v binary_prefix $RELEASE_BINARY_PREFIX_FMT $bucket_name
 
-  # Copy across the binaries and set as latest
+  # Copy across the binaries
   aws s3 cp "$ALPHA_LINUX_BINARY_PATH" "$(printf $RELEASE_LINUX_BINARY_PATH_FMT $binary_prefix $CODEBUILD_RESOLVED_SOURCE_VERSION)" $PUBLIC_CP_ARGS
-  aws s3 cp "$ALPHA_LINUX_BINARY_PATH" "$(printf $RELEASE_LINUX_BINARY_PATH_FMT $binary_prefix latest)" $PUBLIC_CP_ARGS
-
   aws s3 cp "$ALPHA_DARWIN_BINARY_PATH" "$(printf $RELEASE_DARWIN_BINARY_PATH_FMT $binary_prefix $CODEBUILD_RESOLVED_SOURCE_VERSION)" $PUBLIC_CP_ARGS
-  aws s3 cp "$ALPHA_DARWIN_BINARY_PATH" "$(printf $RELEASE_DARWIN_BINARY_PATH_FMT $binary_prefix latest)" $PUBLIC_CP_ARGS
```
cadedaniel: We weren't depending on this before?

RedbackThomson: Just checked all the binary code paths and any copying is done from the commit SHA and not from latest.

codebuild/scripts/package_operators.sh
gautamkmr: +1

gautamkmr:
Changes LGTM.

Can we add some details on PR template. More specifically 
```
Which issue(s) does this PR fix?
Fixes # N/A
```

RedbackThomson:
> Changes LGTM.
> 
> Can we add some details on PR template. More specifically
> 
> ```
> Which issue(s) does this PR fix?
> Fixes # N/A
> ```

We don't have a Github issue associated with this change. I have posted all the information about the PR in the description, instead.

RedbackThomson:
I have reverted the deletion of the push to `latest` - we will continue to do so, but put out documentation clarifying what each tag means and the dangers that comes with it.

PR #92, opened 2020-02-10
Added TrainingJob Example for Complex AugmentedManifest Image Classification

raminetinati:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #97, opened 2020-03-04
Added TrainingJob Example for Complex AugmentedManifest Image Classif…

RedbackThomson:

Fixes #92 

Duplicate of #92 with changes.

I am not sure where this particular sample comes from or the data set it is associated with. If we can figure this out I will include it in a header comment.

### All Submissions:

### DIFF
--- a/samples/image-classifier-augmented-manifest.yaml
+++ b/samples/image-classifier-augmented-manifest.yaml
@@ -1,0 +1,61 @@
+apiVersion: sagemaker.aws.amazon.com/v1
+kind: TrainingJob
+metadata:
+  name: augmented-s3-manifest-image-classifier
+spec:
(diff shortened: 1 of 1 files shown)


samples/image-classifier-augmented-manifest.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: sagemaker.aws.amazon.com/v1
```
gautamkmr: Can we link here more details on this example ?
For instance you can add tutorial, blog or any other useful doc as comment for reference ?

RedbackThomson: I don't have one, sadly. I'm not sure where this example comes from. I did some research online but couldn't nail down any sources that matched.

PR #98, opened 2020-03-05
Sagemaker debugger support in training operator 

gautamkmr:
This PR adds sagemaker debugger support in operator.

N/A

Fixes # 
N/A

N/A

Yes, we need to add example and guideline how to run and explore the debugger job. 

### All Submissions:


gautamkmr:
`make test` runs green on local environment.

tests/codebuild/create_tests.sh
```diff
@@ -18,6 +18,7 @@ function run_canary_tests
   yq w -i testfiles/xgboost-mnist-batchtransform.yaml "spec.modelName" "$(get_sagemaker_model_from_k8s_model xgboost-model)"
   run_test testfiles/xgboost-mnist-batchtransform.yaml 
   run_test testfiles/xgboost-hosting-deployment.yaml
+  run_test testfiles/xgboost-mnist-trainingjob-debugger.yaml
```
RedbackThomson: Also need to add the file to `inject_tests` before it can be used (and replace all bucket names and role ARNs with placeholders)

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
 		break
 
 	case sagemaker.TrainingJobStatusCompleted:
-		if err = r.addModelPathToStatus(ctx); err != nil {
-			return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add model path to status"))
+		if len(ctx.TrainingJob.Status.ModelPath) == 0 {
```
cadedaniel: Thanks for adding this optimization. I think we need to do more than check if the string is empty, though. We should strive to support full eventual consistency, such that if the model path is incorrect we update it. 

We can either update it every time (as it was before) or modify `addModelPathToStatus` so that it only updates if the status model path is different than the actual model path.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+			if err = r.addModelPathToStatus(ctx); err != nil {
+				return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add model path to status"))
+			}
+		}
+		// debug rule runs asynchronously with trainingjob. In some cases debug rule
```
cadedaniel: nit: Can we start off comments with proper sentence grammar? Things like capitalizing the first word helps readability a lot.
```
Debug rules run asynchronously with training jobs. (...)
```

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -99,6 +99,12 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	switch ctx.TrainingJob.Status.TrainingJobStatus {
 	case string(sagemaker.TrainingJobStatusCompleted):
+		// If there are any debug rule job still in progress
```
cadedaniel: nit: Can we flesh out this comment? `If there are any debug rule jobs still in progress, keep requeueing.`

api/v1/common/sagemaker_api.go
```diff
@@ -480,3 +480,51 @@ type TransformResources struct {
...
+	S3OutputPath *string `json:"s3OutputPath"`
+}
+
+// DebugRuleEvalutionStatus https://docs.aws.amazon.com/sagemaker/latest/dg/API_DebugRuleEvaluationStatus.html
+type DebugRuleEvaluationStatus struct {
+	LastModifiedTime *metav1.Time `json:"lastModifiedTime,omitempty"`
+
+	RuleConfigurationName *string `json:"ruleConfigurationName,omitempty"`
+
+	RuleEvaluationJobArn *string `json:"ruleEvaluationJobArn,omitempty"`
```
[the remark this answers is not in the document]
gautamkmr: thats right, for status we don't do any.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -99,6 +99,12 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		for _, debugRuleJob := range ctx.TrainingJob.Status.DebugRuleEvaluationStatuses {
+			if *(debugRuleJob.RuleEvaluationStatus) == string(sagemaker.RuleEvaluationStatusInProgress) || *(debugRuleJob.RuleEvaluationStatus) == string(sagemaker.RuleEvaluationStatusStopping) {
+				return controllers.RequeueAfterInterval(r.PollInterval, nil)
+			}
+		}
 		fallthrough
 	case string(sagemaker.TrainingJobStatusFailed):
```
cadedaniel: Won't we want to continue requeuing if there are debug rule jobs still running in this case? Same for `TrainingJobStatusStopped`.

api/v1/common/sagemaker_api.go
```diff
@@ -480,3 +480,51 @@ type TransformResources struct {
...
+	RuleParameters []*KeyValuePair `json:"ruleParameters,omitempty"`
+}
+
+// DebugHookConfig https://docs.aws.amazon.com/sagemaker/latest/dg/API_DebugHookConfig.html
+type DebugHookConfig struct {
+	LocalPath *string `json:"localPath,omitempty"`
+	// +kubebuilder:validation:Pattern=^(https|s3)://([^/]+)/?(.*)$
+	S3OutputPath             *string                    `json:"s3OutputPath"`
+	HookParameters           []*KeyValuePair            `json:"ruleParameters,omitempty"`
+	CollectionConfigurations []*CollectionConfiguration `json:"collectionConfigurations,omitempty"`
```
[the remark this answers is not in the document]
gautamkmr: Can you point me any reference ? I did not see "20" anywhere in sdk related to CollectionConfigurations

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
-			return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add model path to status"))
+		if len(ctx.TrainingJob.Status.ModelPath) == 0 {
+			if err = r.addModelPathToStatus(ctx); err != nil {
+				return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add model path to status"))
+			}
+		}
+		// debug rule runs asynchronously with trainingjob. In some cases debug rule
+		// lags behind the trainingjob. Hence Training has to keep reconciling until
+		// its debug rule has completed.
+		// This breaks a bit k8s experience, since describe output will change evel
```
cadedaniel: nit: `s/evel/even`.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+		// if trainingjob has been completed.
+		// Its hard to populate the debug rules in `kubectl get` since not every trainifjccinlfg
+		// job will have debug rule and  column prints are not supported conditionally.
+		for _, debugRuleJob := range ctx.TrainingJob.Status.DebugRuleEvaluationStatuses {
+			if *(debugRuleJob.RuleEvaluationStatus) == string(sagemaker.RuleEvaluationStatusInProgress) || *(debugRuleJob.RuleEvaluationStatus) == string(sagemaker.RuleEvaluationStatusStopping) {
```
cadedaniel: What happens if `RuleEvaluationStatus` is nil? We should not crash the operator if it is nil.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+				if err = r.addDebugRuleEvaluationStatusesToStatus(ctx); err != nil {
```
cadedaniel: What happens if multiple `RuleEvaluationStatus` are `InProgress` or `Stopping`? This will generate N updates for N such rules. We should fix to stop looping in that case.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+					return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add debug statuses job to status"))
```
cadedaniel: nit: This error message doesn't make sense, consider "Unable to add debug job statuses to status".

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+			}
+		}
+		// debug rule runs asynchronously with trainingjob. In some cases debug rule
+		// lags behind the trainingjob. Hence Training has to keep reconciling until
+		// its debug rule has completed.
+		// This breaks a bit k8s experience, since describe output will change evel
+		// if trainingjob has been completed.
+		// Its hard to populate the debug rules in `kubectl get` since not every trainifjccinlfg
+		// job will have debug rule and  column prints are not supported conditionally.
+		for _, debugRuleJob := range ctx.TrainingJob.Status.DebugRuleEvaluationStatuses {
```
cadedaniel: I think we'll want to update these statuses when the TrainingJob is InProgress as well. Why are we only updating debug rule statuses when TrainingJob is completed?

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -298,6 +321,19 @@ func (r *Reconciler) removeFinalizer(ctx reconcileRequestContext) error {
 	return nil
 }
 
+// Add information regarding the debugging statuses to the status fields.
+func (r *Reconciler) addDebugRuleEvaluationStatusesToStatus(ctx reconcileRequestContext) error {
```
cadedaniel: This function is named the same way as `addModelPathToStatus` but does something completely different -- it does not call `r.Status().Update()`. We need to be consistent or be excruciatingly clear about the difference.. I don't see any reason for this function to not call `r.Status.Update()`.

api/v1/common/sagemaker_api.go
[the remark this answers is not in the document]
cadedaniel: He's referring to the [actual API](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DebugHookConfig.html#sagemaker-Type-DebugHookConfig-CollectionConfigurations).

That said, I am a bit hesitant to add strict SageMaker validation to our specs. I don't think that changes they make to validation like this will be communicated to us. I think we should rely on our `additional` field if the customer supplies something that doesn't fit SaegMaker validation?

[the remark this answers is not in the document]
gautamkmr: +1 thats what additional fields are for. 
Resolving this.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -185,8 +191,24 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+			if err = r.addModelPathToStatus(ctx); err != nil {
+				return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add model path to status"))
+			}
+		}
+		// debug rule runs asynchronously with trainingjob. In some cases debug rule
+		// lags behind the trainingjob. Hence Training has to keep reconciling until
+		// its debug rule has completed.
+		// This breaks a bit k8s experience, since describe output will change evel
+		// if trainingjob has been completed.
+		// Its hard to populate the debug rules in `kubectl get` since not every trainifjccinlfg
```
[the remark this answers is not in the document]
gautamkmr: My yubi key

controllers/sdkutil/spec_sdk_converters.go
```diff
@@ -106,28 +106,43 @@ func CreateTrainingJobSpecFromDescription(description sagemaker.DescribeTraining
...
+	debugHookConfigCollectionsConfigurationsCollectionParameters := [][]*commonv1.KeyValuePair{}
+
+	if spec.DebugHookConfig != nil {
+		debugHookConfigHookParameters = spec.DebugHookConfig.HookParameters
+		spec.DebugHookConfig.HookParameters = []*commonv1.KeyValuePair{}
+
+		for _, debugHookConfigCollectionConfiguration := range spec.DebugHookConfig.CollectionConfigurations {
+			debugHookConfigCollectionConfigurationCollectionParameters := debugHookConfigCollectionConfiguration.CollectionParameters
+			debugHookConfigCollectionsConfigurationsCollectionParameters = append(debugHookConfigCollectionsConfigurationsCollectionParameters, debugHookConfigCollectionConfigurationCollectionParameters)
+			debugHookConfigCollectionConfiguration.CollectionParameters = []*commonv1.KeyValuePair{}
```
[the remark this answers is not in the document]
gautamkmr: ok, will make the change in new revision.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -99,6 +99,12 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	switch ctx.TrainingJob.Status.TrainingJobStatus {
 	case string(sagemaker.TrainingJobStatusCompleted):
+		// If there are any debug rule job still in progress
+		for _, debugRuleJob := range ctx.TrainingJob.Status.DebugRuleEvaluationStatuses {
+			if *(debugRuleJob.RuleEvaluationStatus) == string(sagemaker.RuleEvaluationStatusInProgress) || *(debugRuleJob.RuleEvaluationStatus) == string(sagemaker.RuleEvaluationStatusStopping) {
```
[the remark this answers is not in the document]
gautamkmr: Thanks for pointing out, TBH I was not very comfortable about this but glad that you point it out.

controllers/aws_config_loader.go
```diff
@@ -61,7 +63,7 @@ func (l *AwsConfigLoader) InstallCredsUsingSDKV1(config *aws.Config, regionOverr
 	// Reference: https://eksctl.io/usage/iamserviceaccounts/
 	if (len(awsAccessKeyId) == 0 || len(awsSecretAccessKey) == 0) && len(awsWebIdentityTokenFile) != 0 && len(awsRoleArn) != 0 {
 		credsProvider := aws.SafeCredentialsProvider{}
-		credsProvider.RetrieveFn = func() (aws.Credentials, error) {
+		credsProvider.RetrieveFn = func(context.Context) (aws.Credentials, error) {
```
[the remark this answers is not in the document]
gautamkmr: I've added a task to remove the reference code related to IAM. I have not created an image and tested it out yet.

api/v1/trainingjob/trainingjob_types.go
```diff
@@ -96,6 +102,8 @@ type TrainingJobStatus struct {
 
 	//Full path to the training artifact (model)
 	ModelPath string `json:"modelPath,omitempty"`
+
+	DebugRuleEvaluationStatuses []commonv1.DebugRuleEvaluationStatus `json:"debugRuleEvaluationStatuses,omitempty"`
```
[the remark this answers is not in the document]
gautamkmr: done.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: Can you please tell me why operator will crash if `RuleEvaluationStatus` is `nil`? This code does not dereference it right?

controllers/sdkutil/spec_sdk_converters.go
```diff
@@ -123,11 +123,36 @@ func CreateCreateTrainingJobInputFromSpec(spec trainingjobv1.TrainingJobSpec) sa
 // This way can be acceptable if we have test coverage that assures breakage when sdk / trainingjobv1.structs diverage.
 // Alternatives: https://quip-amazon.com/3PVUAsbL9I69/how-do-we-convert-between-structs-coming-from-etcd-to-structs-going-to-sagemaker
 func createCreateTrainingJobInputFromSpec(spec trainingjobv1.TrainingJobSpec) (sagemaker.CreateTrainingJobInput, error) {
-
 	var output sagemaker.CreateTrainingJobInput
+
+	// Convert each of the KeyValuePairs manually
```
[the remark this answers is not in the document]
gautamkmr: +1, it was a copy-pasta.

controllers/sdkutil/spec_sdk_converters.go
```diff
@@ -202,6 +240,26 @@ func ConvertHyperParameterTrainingJobSummaryFromSageMaker(source *sagemaker.Hype
...
+
+	for _, status := range source {
+		var target commonv1.DebugRuleEvaluationStatus
+
+		str, err := json.Marshal(status)
+		if err != nil {
+			return nil, err
+		}
+
+		json.Unmarshal(str, &target)
```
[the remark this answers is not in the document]
gautamkmr: I think it was copy pasta too.

controllers/trainingjob/trainingjob_controller.go
cadedaniel: Yes, this code dereferences `RuleEvaluationStatus`. The explicit `*` is what does it.

cadedaniel: Can use `GetOrDefault` in this case.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: Good point, I think we should keep requeuing there as well.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: Right, I overlooked at it. Yes, using `GetOrDefault`

controllers/trainingjob/trainingjob_controller.go
gautamkmr: good point, I think adding `break` helps here.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: done.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: Correct me if I am wrong @RedbackThomson The idea here is every requeue will update status of  debug job as well. So when training job is in progress the debug job status will be updated too, since requeue will execute that code path (I should link here which code path ?) as well, which will not happen if the training job has stopped.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: Agree.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: I will remove the optimization for now and later worry about how to filter update if the path differs.

gautamkmr: Also when do u think mode path will be non-empty and in-correct ?

gautamkmr:
> Looks good overall.
> 
> Can we add an integration test for both 1p hooks and custom hooks - and then have the verify_test step check that they are reported in the statuses?
> 
> Other than that and unit tests, great!
I think example with this PR has 1p hooks. What do you mean by custom hooks ?  There are only few [examples](https://github.com/awslabs/amazon-sagemaker-examples/tree/master/sagemaker-debugger), I got that from a DS worked on this. If you want any specific example from that list let me know I can add that here too.

controllers/trainingjob/trainingjob_controller.go
RedbackThomson: Please correct me if I'm wrong @cadedaniel, but how I interpreted this is that we should be requeue-ing the training job if there are debug rules in progress, regardless of the state of the training job. It's possible that the training job has stopped, but the debug rules are still in progress. If we don't requeue, then the status of those debug rules will always appear "InProgress" in k8s.

controllers/trainingjob/trainingjob_controller.go
RedbackThomson: This code is only executed if the job status is `Completed`. For `InProgress` we would need to add it after the switch block. The reason I put `ModelPath` in here is because that field is only present when the status is `Completed`.

controllers/trainingjob/trainingjob_controller.go
cadedaniel: I don't know when it will be incorrect, but defensive code is better for maintainability.

controllers/trainingjob/trainingjob_controller.go
cadedaniel: I think that the code that updates the debug `RuleEvaluationStatus` should always execute if the TrainingJob exists, regardless of TrainingJob status. So, before the switch statement. I can't think of any good reason to not do so.

controllers/sdkutil/spec_sdk_converters.go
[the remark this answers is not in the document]
cadedaniel: I think you forgot to remove this @gautamkmr

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -322,6 +364,12 @@ func (r *Reconciler) updateStatusWithAdditional(ctx reconcileRequestContext, tra
 	jobStatus.SecondaryStatus = trainingJobSecondaryStatus
 	jobStatus.Additional = additional
 
+	// Check if we should update debugRuleEvaluationStatus
+	if ctx.TrainingJobDescription != nil && len(ctx.TrainingJobDescription.DebugRuleEvaluationStatuses) > 0 {
```
cadedaniel: Why are we duplicating this functionality? We should have only [one piece of code responsible](https://en.wikipedia.org/wiki/Single_responsibility_principle) for updating the debug rule statuses.

api/v1/trainingjob/trainingjob_types.go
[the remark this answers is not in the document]
cadedaniel: @gautamkmr can we use proper capitalization on documentation like this? When we eventually publish docs, this will be customer facing. `s/debughookconfig/DebugHookConfig`.

I think we can make the sentence a bit more clear, something like 
```go
// Status of rule evaluation jobs, obtained from DebugRuleEvaluationStatuses.
// https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DescribeTrainingJob.html#sagemaker-DescribeTrainingJob-response-DebugRuleEvaluationStatuses
```

controllers/sdkutil/spec_sdk_converters.go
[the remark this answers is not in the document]
gautamkmr: ops yeah

controllers/trainingjob/trainingjob_controller.go
gautamkmr: May be I am not communicating it properly. We do not update status of debug job asynchronously but every requeue does it automatically. In case of  training job `completion` we need to check the debug job to make sure it also reaches to its final state, but when training job is in progress requeue cycle will anyway update the debug job. So what is the issue here?

gautamkmr: Also since Trainingjob and debug job runs asynchronously I noticed in sagemaker even if you explicitly stop a trainingjob, debug job will stay in progress.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: I think `updateStatusWithAdditional` update the status of object. In this function if you notice we populate all the trainingjob attribute and update it, hence the debugrule status

gautamkmr: Fixed

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -754,6 +1042,25 @@ func CreateDescribeOutputWithOnlyStatus(status sagemaker.TrainingJobStatus, seco
...
+		DebugRuleEvaluationStatuses: []sagemaker.DebugRuleEvaluationStatus{
+			{
+				RuleEvaluationStatus: sagemaker.RuleEvaluationStatusIssuesFound,
+			},
+		},
+	}
+}
+
+// Helper function to create a DescribeTrainingJobOutput with debug job status
+func CreateDescribeOutputWithDebugJobStatus(status sagemaker.TrainingJobStatus, secondaryStatus sagemaker.SecondaryStatus, debugJobstatus sagemaker.RuleEvaluationStatus) sagemaker.DescribeTrainingJobOutput {
```
cadedaniel: nit: typo `s/debugJobstatus/debugJobStatus/`

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -754,6 +1042,25 @@ func CreateDescribeOutputWithOnlyStatus(status sagemaker.TrainingJobStatus, seco
...
+	return sagemaker.DescribeTrainingJobOutput{
+		TrainingJobStatus: status,
+		SecondaryStatus:   secondaryStatus,
```
cadedaniel: nit: Can use the other helper function to reduce code duplication:

```go
func CreateDescribeOutputWithDebugJobStatus(status sagemaker.TrainingJobStatus, secondaryStatus.SecondaryStatus, debugJobStatus sagemaker.RuleEvaluationStatus) sagemaker.DescribeTrainingJobOutput {
    output := CreateDescribeOutputWithOnlyStatus(status, secondaryStatus)
    output.DebugRuleEvaluationStatuses = []sagemaker.DebugRuleEvaluationStatus{
        {
            RuleEvaluationStatus: debugJobStatus,
        },
    },
}
```

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,294 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+			})
+
+			When("!HasDeletionTimestamp", func() {
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+
+				It("Updates status", func() {
+					ExpectStatusToBe(trainingJob, string(expectedStatus), string(expectedSecondaryStatus))
+				})
```
cadedaniel: Why are we not testing that the `debugStatus` is correctly set?

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -754,6 +1042,25 @@ func CreateDescribeOutputWithOnlyStatus(status sagemaker.TrainingJobStatus, seco
 	return sagemaker.DescribeTrainingJobOutput{
 		TrainingJobStatus: status,
 		SecondaryStatus:   secondaryStatus,
+		// Added debug job completion status for cases when debug job has finished
+		DebugRuleEvaluationStatuses: []sagemaker.DebugRuleEvaluationStatus{
+			{
+				RuleEvaluationStatus: sagemaker.RuleEvaluationStatusIssuesFound,
+			},
+		},
+	}
```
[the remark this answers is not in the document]
gautamkmr: I wanted to test the case when debug job is finished before trainingjob. I'll rename it.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,294 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+				})
+			})
+
+			When("HasDeletionTimestamp", func() {
+				BeforeEach(func() {
+					shouldHaveDeletionTimestamp = true
+				})
+
+				It("Deletes the training job", func() {
+					ExpectTrainingJobToBeDeleted(trainingJob)
```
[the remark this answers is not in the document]
gautamkmr: `Is this right? When the training job status is "Failed" and the debug job status is "Stopping", we will allow the Kubernetes job to be deleted? `
There is not k8s job for debug job, so we are not deleting anything. The debug job can be in stopping state due to various reasons. 

`Do the debug jobs charge the user while they run? If so I think we should maybe wait until the DebugJobs are finished before we delete the etcd entry. Whatever we do here, we should do for all other HasDeletionTimestamp debug job cases.`

I don't know about pricing, couldn't find any reference. I don't think we should do that way, ideally debugger should have been designed to be like a spawner job (similar to HPO). With current state I am a bit hesitant to go in that direction.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -170,6 +177,21 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+	// lags behind the trainingjob. Hence Training has to keep reconciling until
+	// its debug rule has completed.
+	// This breaks a bit k8s experience, since describe output will change even
+	// if trainingjob might have been completed.
+	// Its hard to populate the debug rules in `kubectl get` since not every trainingjob
+	// job will have debug rule and  column prints are not supported conditionally.
+	// Update the status of all the debug jobs
+	// TODO: Optimize the update generation
+	for _, _ = range ctx.TrainingJobDescription.DebugRuleEvaluationStatuses {
+		if err = r.addDebugRuleEvaluationStatusesToStatus(ctx); err != nil {
```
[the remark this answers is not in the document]
gautamkmr: I think you just gave me idea :)

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -170,6 +177,21 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
 
+	// Debug rule runs asynchronously with trainingjob. In some cases debug rule
+	// lags behind the trainingjob. Hence Training has to keep reconciling until
+	// its debug rule has completed.
+	// This breaks a bit k8s experience, since describe output will change even
+	// if trainingjob might have been completed.
+	// Its hard to populate the debug rules in `kubectl get` since not every trainingjob
+	// job will have debug rule and  column prints are not supported conditionally.
+	// Update the status of all the debug jobs
+	// TODO: Optimize the update generation
```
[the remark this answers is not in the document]
gautamkmr: Will fix it.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -126,6 +132,7 @@ type reconcileRequestContext struct {
 }
 
 func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
+	ctx.Log.Info("reconcileTrainingJob() started")
```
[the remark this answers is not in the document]
gautamkmr: done.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,294 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+				})
+
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+			})
+		})
+
+		Context("TrainingJob with debug job is `stopping` has status 'Failed'", func() {
```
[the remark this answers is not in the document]
gautamkmr: done.

controllers/trainingjob/trainingjob_controller_test.go
gautamkmr: Added the test for it, thanks for pointing this out.

controllers/trainingjob/trainingjob_controller_test.go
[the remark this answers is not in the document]
gautamkmr: TBH if person is using this function he/she should read it well :) ;)

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -170,6 +176,25 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
 		}
 	}
 
+	// Check  sagemaker describe output for each debug jobs and
+	// generates the update if k8s status for corresponding debug jobs differ.
+	// Initially status will be empty so we will generate update for all debug jobs
+	if len(ctx.TrainingJobDescription.DebugRuleEvaluationStatuses) != len(ctx.TrainingJob.Status.DebugRuleEvaluationStatuses) {
+		for _, _ = range ctx.TrainingJobDescription.DebugRuleEvaluationStatuses {
+			if err = r.addDebugRuleEvaluationStatusesToStatus(ctx); err != nil {
```
RedbackThomson: Does this loop run `addDebugRuleEvaluationStatusesToStatus` for every status? Wouldn't this be adding them a total of `len(ctx.TrainingJobDescription.DebugRuleEvaluationStatuses)` times? From our definition of that method, it adds all of them at once, so we only need to run it once.

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -170,6 +176,25 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
+				return r.updateStatusAndReturnError(ctx, ReconcilingTrainingJobStatus, "", errors.Wrap(err, "Unable to add debug job statuses to status"))
+			}
+		}
+	} else {
+		for i, debugJob := range ctx.TrainingJobDescription.DebugRuleEvaluationStatuses {
+			if string(debugJob.RuleEvaluationStatus) != controllers.GetOrDefault(ctx.TrainingJob.Status.DebugRuleEvaluationStatuses[i].RuleEvaluationStatus, "") {
+				if err = r.addDebugRuleEvaluationStatusesToStatus(ctx); err != nil {
```
RedbackThomson: Same as above, if `addDebugRuleEvaluationStatusesToStatus` adds all the rule evaluation statuses to the status then we only need to run it once (break out of the loop).

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,296 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+					AddDescribeTrainingJobResponse(describeOutput)
+			})
+
+			When("!HasDeletionTimestamp", func() {
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+
+				It("Updates status", func() {
+					ExpectJobsStatusToBe(trainingJob, string(expectedStatus), string(expectedSecondaryStatus), string(debugJobStatus))
```
RedbackThomson: Should we create a second helper function `ExpectDebugJobStatusToBe` so that we don't need to bother unit testing the debug statuses in every other test?

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,296 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+					ExpectTrainingJobToBeDeleted(trainingJob)
+				})
+
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+			})
+		})
+
+		Context("TrainingJob is 'Failed' and has debug job with status 'Stopping'", func() {
```
RedbackThomson: For consistency with the other contexts can we make this `TrainingJob is 'Failed' and has debug status 'Stopping'`

controllers/trainingjob/trainingjob_controller_test.go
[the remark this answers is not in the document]
RedbackThomson: I agree with @gautamkmr here in that our job was to manage the training job that they started, and if they attach a debugger then they assume any costs incurred to that configuration. If the user sees their TrainingJob as "Failed" then they would assume it would be possible to delete it - it may be unclear as to why we are restricting this ability.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,296 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+					ExpectTrainingJobToBeDeleted(trainingJob)
+				})
+
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+			})
+		})
+
+		Context("TrainingJob is 'Stopped' and has debug job with status 'InProgress'.", func() {
```
RedbackThomson: nit: Consistency, see above.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: I see.

controllers/trainingjob/trainingjob_controller.go
gautamkmr: In that case I'll just un-conditionally call that method.

controllers/trainingjob/trainingjob_controller_test.go
gautamkmr: I thought that way but then code duplications remains same.

gautamkmr: I've renamed, hope that will be clear.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,301 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+					AddDescribeTrainingJobResponse(describeOutput)
+			})
+
+			When("!HasDeletionTimestamp", func() {
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+
+				It("Updates status", func() {
+					ExpectAllJobsStatusToBe(trainingJob, string(expectedStatus), string(expectedSecondaryStatus), debugJobStatus)
```
cadedaniel: `ExpectAllJobsStatusToBe` name is confusing, can do either `ExpectStatusAndDebugStatusToBe` or separate `ExpectStatusToBe` and `ExpectDebugJobStatusesToBe`.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -726,6 +1021,23 @@ func ExpectStatusToBe(trainingJob *trainingjobv1.TrainingJob, primaryStatus, sec
...
+// Expect trainingjob status and debug job status to be as specified
+func ExpectAllJobsStatusToBe(trainingJob *trainingjobv1.TrainingJob, primaryStatus, secondaryStatus string, debugJobStatus []sagemaker.RuleEvaluationStatus) {
+	var actual trainingjobv1.TrainingJob
+	err := k8sClient.Get(context.Background(), types.NamespacedName{
+		Namespace: trainingJob.ObjectMeta.Namespace,
+		Name:      trainingJob.ObjectMeta.Name,
+	}, &actual)
+	Expect(err).ToNot(HaveOccurred())
+
+	Expect(string(actual.Status.TrainingJobStatus)).To(Equal(primaryStatus))
```
cadedaniel: nit: This `string` conversion shouldn't be necessary, the source is a string.

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -726,6 +1021,23 @@ func ExpectStatusToBe(trainingJob *trainingjobv1.TrainingJob, primaryStatus, sec
...
+	Expect(string(actual.Status.SecondaryStatus)).To(Equal(secondaryStatus))
+
+	for i, _ := range debugJobStatus {
+		Expect(string(*actual.Status.DebugRuleEvaluationStatuses[i].RuleEvaluationStatus)).To(Equal(string(debugJobStatus[i])))
```
cadedaniel: nit: might be able to use this:
```go
for i, debugJobStatus := range debugJobStatuses {
    Expect(A[i]).To(Equal(string(debugJobStatus)))
}
```

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,301 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+			})
+		})
+
+		Context("TrainingJob is 'Stopped' and has debug job with status 'InProgress'.", func() {
+			var debugJobStatus []sagemaker.RuleEvaluationStatus
+			BeforeEach(func() {
+				expectedStatus = sagemaker.TrainingJobStatusStopped
+				debugJobStatus = append(debugJobStatus, sagemaker.RuleEvaluationStatusInProgress)
```
cadedaniel: nit: can use `debugJobStatus = type_here[]{sagemaker.RuleEvaluationStatusInProgress, sagemaker.RuleEvaluationStatusIssuesFound}`

controllers/trainingjob/trainingjob_controller.go
RedbackThomson: Bumping this comment. Seems like we still have unnecessary looping.

controllers/trainingjob/trainingjob_controller_test.go
RedbackThomson: ++ on separate still

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -617,6 +617,301 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
...
+					ExpectTrainingJobToBeDeleted(trainingJob)
+				})
+
+				It("Requeues after interval", func() {
+					ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				})
+			})
+		})
+
+		Context("TrainingJob is 'Stopped' and has debug job with status 'InProgress'.", func() {
```
RedbackThomson: nit: Update to match multiple debug statuses

controllers/trainingjob/trainingjob_controller.go
gautamkmr: I think you missed the `break` in line 192. the loop is necessary to find the job where k8s status and sm status are different.

PR #104, opened 2020-03-18
Namespaced installer Kustomize targets

RedbackThomson:
I have added a set of Kustomize targets to support our new namespaced installation method. These require duplicating the RBAC resources to fit within the namespace scope. As a result, users cannot, by default, submit jobs through `kubectl apply -f -` since that user is by default authenticated as `system:default`. Instead, users will have to authenticate with a ServiceAccount with proper privileges. I can provide samples for how this should be done.

I did a diff between the existing installer and the one produced after this PR change and they are identical. I have also tested my installer against the hand crafted ones produced by @mbaijal and they are also identical.

Fixes #

### All Submissions:


config/installers/rolebasedcreds/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,66 @@
...
+# Rename created namespace
+- target:
+    version: v1
+    kind: Namespace
+    name: sagemaker-k8s-operator-system
+  patch: |-
+    - op: replace
+      path: "/metadata/name"
+      value: PLACEHOLDER-NAMESPACE
+# Patch broken ServiceAccount names/namespaces (they all default back to `default`)
```
cadedaniel: Can add comment explaining that the namespaces are set via line 5, whereas the names are reset to `default` and so must be set manually? By my reading I believe that is what is happening here. Perhaps even add a link to whatever Github issue you found so that we can remove this in future if Kustomize ever fixes it.

Nice fix.

config/rbac/namespaced/role.yaml
```diff
@@ -0,0 +1,168 @@
+
+---
```
cadedaniel: nit: remove

config/rbac/namespaced/role.yaml
```diff
@@ -0,0 +1,168 @@
...
+  - batchtransformjobs
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
+- apiGroups:
```
cadedaniel: What is our plan to keep these updated? Normally, kubebuilder generators produce new CRD roles [from annotation](https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/controllers/trainingjob/trainingjob_controller.go#L72-L73) so we don't have to worry about it.

We should have either automated way to do it or documentation, this will have to be done for any new k8s resource. If you don't already have a better way to do it, I think the generators can be configured to write to this Kustomize overlay dependency, or namespaced scope kustomize overlays can depend on normal CRD overlay (which is already setup with generators).

RedbackThomson: Yeah I might have to write patches for things that are auto-generated by kubebuilder. I will run the generator and amend anything it produces - problem may be with the default paths it is installed into. Kustomize requires resources exist at the same level or deeper (which really messes with how I can organise the folders).

RedbackThomson:
> Nice work! Can we test the new+existing installers and make sure they work as expected?

I did a diff between the existing installer and the one produced after this PR change and they are identical. I have also tested my installer against the hand crafted ones produced by @mbaijal and they are also identical.

cadedaniel:
> > Nice work! Can we test the new+existing installers and make sure they work as expected?
> 
> I did a diff between the existing installer and the one produced after this PR change and they are identical. I have also tested my installer against the hand crafted ones produced by @mbaijal and they are also identical.

Ok, awesome. Maybe state this in a "Testing" section of PR so I don't have to ask you to explain :)

config/installers/rolebasedcreds/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+patchesStrategicMerge:
+- resources/manager_add_service_account_patch.yaml
+
+patchesJson6902:
+# Rename created namespace
+- target:
+    version: v1
+    kind: Namespace
+    name: sagemaker-k8s-operator-system
```
gautamkmr: Q: So is this the default namespace and gets overwritten if user specifies it ?

config/rbac/namespaced/role.yaml
cadedaniel: Ok, wow that is an annoying constraint by Kustomize. Ideally we don't have to do anything manual to do CRD updates, hopefully that's easy to implement

config/installers/rolebasedcreds/namespaced/kustomization.yaml
gautamkmr: ++ on comment for explanation.

config/installers/rolebasedcreds/namespaced/kustomization.yaml
RedbackThomson: This block fetches the namespace that is created by the default Kustomize target and renames it to "PLACEHOLDER-NAMESPACE". We are planning on creating a script that replaces all instances of "PLACEHOLDER-NAMESPACE" with the desired namespace of the installation.

config/rbac/namespaced/role.yaml
gautamkmr: ok, this is a bit a repelling,  if we have to manually update the CRD then it will cause issue in future. For developer that experience might be not good. what all options we have here ?

config/installers/rolebasedcreds/namespaced/kustomization.yaml
gautamkmr: Where is default kustomize target ?My question is more towards what happens if user does not specifies the namespace?

RedbackThomson: The user never needs to "specify" a namespace. They either use the cluster-spaced installer that we have provided previously, or they can modify this namespaced installer (replacing the string) to install into their specific namespace. If they install this without modifying it, it will throw an error because a namespace must have all lowercase letters.

RedbackThomson: > Where is default kustomize target ?

`config/installers/rolebasedcreds/kustomization.yaml`

config/rbac/namespaced/role.yaml
RedbackThomson: It's generated this way, believe it or not. (I believe it)

config/rbac/namespaced/role.yaml
RedbackThomson: Turns out it actually wasn't too hard to change. Just had to manage my directories a little bit, but I removed code duplication fairly easily and just created a patch to change ClusterRole to Role when generating the namespace scoped installers.

Makefile
```diff
@@ -100,4 +100,6 @@ CONTROLLER_GEN=$(shell which controller-gen)
 endif
 
 create-installers: set-image
-	kustomize build config/rolebasedcreds > release/rolebased/installer.yaml
+	kustomize build config/installers/rolebasedcreds > release/rolebased/installer.yaml
```
mbaijal: This breaks the integ tests

mbaijal: This breaks integ tests

gautamkmr: is it because of RBAC role separation ?

RedbackThomson: We don't use these YAML files to install it during integration tests. It's because I removed RBAC and CRDs from the default base, but didn't add it as a dependency back to the default creds installer.

config/bindings/clusterspaced/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+resources:
```
cadedaniel: Can we use better name for the Kustomize target? I am referring to `bindings`.  At first glance I thought `bindings` refers to `RoleBinding`, but I see it contains roles, role bindings, and a service. 

It is important to use apt names because it helps readers understand better the purpose of targets. If there simply isn't a good name, it either suggests we should pick better boundaries or add good comments in these `kustomization.yaml` files.

config/installers/rolebasedcreds/kustomization.yaml
```diff
@@ -4,12 +4,14 @@ kind: Kustomization
 # Adds namespace to all resources.
```
cadedaniel: Under this organization, we have the following structure:

```
config
    default
    installers
        rolebased
            namespaced
    ...
```

I think it is better to have:

```
config
    default # later we can remove this or move it into installers
    installers # (all rolebased)
        clusterscope
        namespace
    ...
```
IMO having `namespaced` be a child of `cluster-scope rolebased` is bad because it conveys the incorrect idea that `namespaced` depends on `clusterscope`, which it doesn't. Additionally it is good idea to avoid unnecessary nesting.

Any reason why `namespaced` needs to be a child of `rolebased`?

This might be better as separate comment, but maybe we should structure release directory same way? `namespace` and `clusterspace` are siblings, one not dependent on the other.

config/default/kustomization.yaml
```diff
@@ -14,4 +14,6 @@ secretGenerator:
 patchesStrategicMerge:
 - manager_creds_patch.yaml
 resources:
+- ../crd
 - ../base
+- ../bindings/clusterspaced
```
cadedaniel: I think in future we should refactor `credsbased/clusterscope`, `credsbased/namespacescope`, `rolebased/clusterscope`, and `rolebased/namespacescope` to depend on two common installers, `clusterscope` and `namespacescope`. They would then patch in their specific authentication mechanisms. This will reduce duplication of dependencies and make edits easier since they have to be done in fewer places.

I'm not necessarily saying we should do this now, just trying to share the little knowledge of Kustomize I have.

config/bindings/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Comment the following 3 lines if you want to disable
+# the auth proxy (https://github.com/brancz/kube-rbac-proxy)
+# which protects your /metrics endpoint.
+- auth_proxy_service.yaml
+- auth_proxy_role.yaml
+- auth_proxy_role_binding.yaml
+
+patchesJson6902:
+# Modify ClusterRole to Role
+- target:
```
cadedaniel: I am curious why we only have to modify one cluster role (`manager-role`), when I know there are four such "Cluster" resources: `manager-role`, `manager-rolebinding`, `proxy-role`, and `proxy-rolebinding`. Why are we taking different approach for `manager-role` vs. the others?

config/bindings/namespaced/auth_proxy_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
cadedaniel: Why are these `proxy-role`, `proxy-rolebinding`, and `manager-rolebinding` duplicated between namespace and clusterscope targets? We should not duplicate these lest we want to invite bugs when we update one and not the other...

config/installers/rolebasedcreds/kustomization.yaml
RedbackThomson: I would love to redesign the structure of these folders, and completely agree it is a little nonsensical at the moment. However I am forced to make silly children folders because Kustomize restricts any resources to be within a child folder of the Kustomization target. Therefore if I want to share resources between two separate targets I have to put one of them into a child folder of the other. If you have better naming I'd happily accept those, though.

config/bindings/namespaced/kustomization.yaml
RedbackThomson: The others, by default, are namespaced to the resources they are created. As long as we specify those resources to be a "Role" and not a "ClusterRole" (which we do in the `bindings` folder) then the bindings namespace properly. The manager role, though, is created by Kubebuilder, so we cannot make a copy of it - this would require engineers to modify the copy each time. Therefore we use a patch to force it to be a namespace Role.

config/bindings/namespaced/auth_proxy_role_binding.yaml
RedbackThomson: I don't know of a good way to remove code duplication here. All the documentation about overlays only modifies the spec of the resources, relying on the Group and Kind as the selectors. We would like to modify the Group and Kind. The only real way to do this is the messy JSON patches. I did it this way before but it made for a lot of fairly unintelligible code and broke the resource resolution that Kustomize does behind the scenes.

PR #108, opened 2020-03-27
Integ Tests for namespace deployment and automated OIDC IAM role generation

mbaijal:
1. Integ Tests for namespace deployment 
2. Automated OIDC IAM role generation for the cluster and given namespace.
3. Some cleanup, but much more possible. 

Todo (in a separate PR):
1. Canaries
2. A better input for the `test-namespace` instead of a hardcoded global variable. 
3. `crd_namespace` is not the most appropriate variable name, change to operator_namespace
4. These are simple tests for namespace deployment, should add some more feature tests. Opinions?
5. updated integ test README.

Fixes #

### All Submissions:

### DIFF
--- a/scripts/generate_iam_role.sh
+++ b/scripts/generate_iam_role.sh
@@ -4,29 +4,29 @@
 # https://sagemaker.readthedocs.io/en/stable/amazon_sagemaker_operators_for_kubernetes.html#create-an-iam-role
 #
 # Run as:
-# $ ./generate_iam_role.sh ${cluster_arn} ${operator_namespace} ${role_name}
+# $ ./generate_iam_role.sh ${cluster_arn/cluster_name} ${operator_namespace} ${role_name} ${cluster_region}
 #
 
-CLUSTER_ARN=${1}
-OPERATOR_NAMESPACE=${2}
-ROLE_NAME=${3}
+CLUSTER_ARN="${1}"
+OPERATOR_NAMESPACE="${2}"
+ROLE_NAME="${3}"
+CLUSTER_REGION="${4:-us-east-1}"
 aws_account=$(aws sts get-caller-identity --query Account --output text)
 trustfile="trust.json"
 
-# Use the cluster arn to get the region and cluster name
+# if using an existing cluster, use the cluster arn to get the region and cluster name
 # example, cluster_arn=arn:aws:eks:us-east-1:12345678910:cluster/test
 cluster_name=$(echo ${CLUSTER_ARN} | cut -d'/' -f2)
-cluster_region=$(echo ${CLUSTER_ARN} | cut -d':' -f4)
 
 # A function to get the OIDC_ID associated with an EKS cluster
 function get_oidc_id {
     # TODO: Ideally this should be based on version compatibility instead of command failure
-    eksctl utils associate-iam-oidc-provider --cluster ${cluster_name} --region ${cluster_region} --approve
+    eksctl utils associate-iam-oidc-provider --cluster ${cluster_name} --region ${CLUSTER_REGION} --approve
     if [[ $? -ge 1 ]]; then
-        eksctl utils associate-iam-oidc-provider --name ${cluster_name} --region ${cluster_region} --approve
+        eksctl utils associate-iam-oidc-provider --name ${cluster_name} --region ${CLUSTER_REGION} --approve
     fi
     
-    local oidc=$(aws eks describe-cluster --name ${cluster_name} --region ${cluster_region} --query cluster.identity.oidc.issuer --output text)
+    local oidc=$(aws eks describe-cluster --name ${cluster_name} --region ${CLUSTER_REGION} --query cluster.identity.oidc.issuer --output text)
     oidc_id=$(echo ${oidc} | rev | cut -d'/' -f1 | rev)
 }
 
@@ -58,7 +58,7 @@ get_oidc_id
 echo "Delete the trust json file if it already exists"
 delete_generated_file "${trustfile}"
 echo "Generate a trust json"
-./generate_trust_policy.sh ${cluster_region} ${aws_account} ${oidc_id} ${OPERATOR_NAMESPACE} > "${trustfile}"
+./generate_trust_policy.sh ${CLUSTER_REGION} ${aws_account} ${oidc_id} ${OPERATOR_NAMESPACE} > "${trustfile}"
 echo "Create the IAM Role using these values"
 create_namespaced_iam_role "${trustfile}"
 echo "Cleanup for the next run!"
--- a/scripts/generate_iam_role.sh
+++ b/scripts/generate_iam_role.sh
@@ -31,7 +31,10 @@ function get_oidc_id {
 }
 
 # A function that generates an IAM role for the given account, cluster, namespace, region
+# Parameter:
+#    $1: Name of the trust file to generate.
 function create_namespaced_iam_role {
+    local trustfile="${1}"
     # Check if role already exists
     aws iam get-role --role-name ${ROLE_NAME}
     if [[ $? -eq 0 ]]; then
@@ -44,18 +47,20 @@ function create_namespaced_iam_role {
 }
 
 # Remove the generated trust file
-function cleanup {
-    rm ${trustfile} 
+# Parameter:
+#    $1: Name of the trust file to delete.
+function delete_generated_file {
+    rm "${1}" 
 }
 
 echo "Get the OIDC ID for the cluster"
 get_oidc_id
 echo "Delete the trust json file if it already exists"
-cleanup
+delete_generated_file "${trustfile}"
 echo "Generate a trust json"
 ./generate_trust_policy.sh ${cluster_region} ${aws_account} ${oidc_id} ${OPERATOR_NAMESPACE} > "${trustfile}"
 echo "Create the IAM Role using these values"
-create_namespaced_iam_role
+create_namespaced_iam_role "${trustfile}"
 echo "Cleanup for the next run!"
-cleanup
+delete_generated_file "${trustfile}"
 
--- a/scripts/generate_iam_role.sh
+++ b/scripts/generate_iam_role.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+# Helper script to generate an IAM Role needed to install operator using role-based authentication.
+# https://sagemaker.readthedocs.io/en/stable/amazon_sagemaker_operators_for_kubernetes.html#create-an-iam-role
+#
+# Run as:
+# $ ./generate_iam_role.sh ${cluster_arn} ${operator_namespace} ${role_name}
+#
+
+CLUSTER_ARN=${1}
+OPERATOR_NAMESPACE=${2}
+ROLE_NAME=${3}
+aws_account=$(aws sts get-caller-identity --query Account --output text)
+trustfile="trust.json"
+
+# Use the cluster arn to get the region and cluster name
+# example, cluster_arn=arn:aws:eks:us-east-1:12345678910:cluster/test
+cluster_name=$(echo ${CLUSTER_ARN} | cut -d'/' -f2)
+cluster_region=$(echo ${CLUSTER_ARN} | cut -d':' -f4)
+
+# A function to get the OIDC_ID associated with an EKS cluster
+function get_oidc_id {
+    # TODO: Ideally this should be based on version compatibility instead of command failure
+    eksctl utils associate-iam-oidc-provider --cluster ${cluster_name} --region ${cluster_region} --approve
+    if [[ $? -ge 1 ]]; then
+        eksctl utils associate-iam-oidc-provider --name ${cluster_name} --region ${cluster_region} --approve
+    fi
+    
+    local oidc=$(aws eks describe-cluster --name ${cluster_name} --region ${cluster_region} --query cluster.identity.oidc.issuer --output text)
+    oidc_id=$(echo ${oidc} | rev | cut -d'/' -f1 | rev)
+}
+
+# A function that generates an IAM role for the given account, cluster, namespace, region
+function create_namespaced_iam_role {
+    # Check if role already exists
+    aws iam get-role --role-name ${ROLE_NAME}
+    if [[ $? -eq 0 ]]; then
+        echo "A role for this cluster and namespace already exists in this account, assuming sagemaker access and proceeding."
+    else
+        echo "IAM Role does not exist, creating a new Role for the cluster"
+        aws iam create-role --role-name ${ROLE_NAME} --assume-role-policy-document file://${trustfile} --output=text
+        aws iam attach-role-policy --role-name ${ROLE_NAME}  --policy-arn arn:aws:iam::aws:policy/AmazonSageMakerFullAccess
+    fi
+}
+
+# Remove the generated trust file
+function cleanup {
+    rm ${trustfile} 
+}
+
+echo "Get the OIDC ID for the cluster"
+get_oidc_id
+echo "Delete the trust json file if it already exists"
+cleanup
+echo "Generate a trust json"
+./generate_trust_policy.sh ${cluster_region} ${aws_account} ${oidc_id} ${OPERATOR_NAMESPACE} > "${trustfile}"
+echo "Create the IAM Role using these values"
+create_namespaced_iam_role
+echo "Cleanup for the next run!"
+cleanup
+
--- a/tests/codebuild/common.sh
+++ b/tests/codebuild/common.sh
@@ -1,42 +1,51 @@
 #!/bin/bash
 
-# Create a resource (run a test) given a specification yaml.
-# Parameter: $1 filename of test
+# Create a resource (run a test) given a specification within a given namespace.
+# Parameter:
+#    $1: Target namespace
+#    $2: Filename of the test
 function run_test()
 {
-  kubectl apply -f "$1"
+  local target_namespace="$1"
+  local file_name="$2"
+
+  kubectl apply -n "$target_namespace" -f "$file_name"
 }
 
 # This function gets the sagemaker model name from k8s model
 # Parameter:
-#    $1: Name of k8s model
-# e.g. get_sagemaker_model_from_k8s_model k8s_model 
+#    $1: Namespace of the k8s model
+#    $2: Name of k8s model
+# e.g. get_sagemaker_model_from_k8s_model default k8s_model 
 function get_sagemaker_model_from_k8s_model()
 {
-  local k8s_model_name="$1"
+  local model_namespace="$1"
+  local k8s_model_name="$2"
   # Get the second line and 3rd column, since valid output will be following
   # NAME                    STATUS    SAGE-MAKER-MODEL-NAME
   # xgboost-model           Created   model-5c06b18921e411ea91230292f5024981
-  kubectl get model $k8s_model_name | sed -n '2 p' |  awk '{print $3}' 
+  kubectl get -n "${model_namespace}" model "${k8s_model_name}" | sed -n '2 p' |  awk '{print $3}' 
 }
 
 # This function waits until a CRD has reached a particular state, or times out
 # Parameter:
-#    $1: Kind of CRD
-#    $2: Instance of CRD
-#    $3: Timeout to complete the test
-#    $4: The status that verifies the job has succeeded.
+#    $1: Namespace of CRD
+#    $2: Kind of CRD
+#    $3: Instance of CRD
+#    $4: Timeout to complete the test
+#    $5: The status that verifies the job has succeeded.
 function wait_for_crd_status()
 {
-  local crd_type="$1"
-  local crd_instance="$2"
-  local timeout="$3"
-  local desired_status="$4"
+  local crd_namespace="$1"
+  local crd_type="$2"
+  local crd_instance="$3"
+  local timeout="$4"
+  local desired_status="$5"
 
   timeout "${timeout}" bash -c \
-    'until [ "$(kubectl get "$0" "$1" -o=custom-columns=STATUS:.status | grep -i "$2" | wc -l)" -eq "1" ]; do \
+    'until [ "$(kubectl get -n "$0" "$1" "$2" -o=custom-columns=STATUS:.status | grep -i "$3" | wc -l)" -eq "1" ]; do \
       sleep 5; \
-    done' "$crd_type" "$crd_instance" "$desired_status"
+    done' "$crd_namespace" "$crd_type" "$crd_instance" "$desired_status"
 
   if [ $? -ne 0 ]; then
     return 1
--- a/tests/codebuild/common.sh
+++ b/tests/codebuild/common.sh
@@ -55,9 +55,10 @@ function wait_for_crd_status()
 # Cleans up all resources created during tests.
 function delete_all_resources()
 {
-  kubectl delete hyperparametertuningjob --all
-  kubectl delete trainingjob --all
-  kubectl delete batchtransformjob --all
-  kubectl delete hostingdeployment --all
-  kubectl delete model --all
+  local crd_namespace="$1"
+  kubectl delete -n "$crd_namespace" hyperparametertuningjob --all 
+  kubectl delete -n "$crd_namespace" trainingjob --all
+  kubectl delete -n "$crd_namespace" batchtransformjob --all
+  kubectl delete -n "$crd_namespace" hostingdeployment --all 
+  kubectl delete -n "$crd_namespace" model --all 
 }
--- a/tests/codebuild/common.sh
+++ b/tests/codebuild/common.sh
@@ -53,6 +53,8 @@ function wait_for_crd_status()
 }
 
 # Cleans up all resources created during tests.
+# Parameter:
+#    $1: Namespace of CRD
 function delete_all_resources()
 {
   local crd_namespace="$1"
--- a/tests/codebuild/create_tests.sh
+++ b/tests/codebuild/create_tests.sh
@@ -9,16 +9,16 @@ function run_canary_tests
   echo "Running canary tests"
   inject_all_variables
 
-  run_test testfiles/xgboost-mnist-trainingjob.yaml
-  run_test testfiles/xgboost-mnist-hpo.yaml
+  run_test default testfiles/xgboost-mnist-trainingjob.yaml
+  run_test default testfiles/xgboost-mnist-hpo.yaml
   # Special code for batch transform till we fix issue-59
-  run_test testfiles/xgboost-model.yaml
+  run_test default testfiles/xgboost-model.yaml
   # We need to get sagemaker model before running batch transform
-  verify_test Model xgboost-model 1m Created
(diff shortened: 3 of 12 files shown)


scripts/generate_iam_role.sh
```diff
@@ -0,0 +1,61 @@
...
+# example, cluster_arn=arn:aws:eks:us-east-1:12345678910:cluster/test
+cluster_name=$(echo ${CLUSTER_ARN} | cut -d'/' -f2)
+cluster_region=$(echo ${CLUSTER_ARN} | cut -d':' -f4)
+
+# A function to get the OIDC_ID associated with an EKS cluster
+function get_oidc_id {
+    # TODO: Ideally this should be based on version compatibility instead of command failure
+    eksctl utils associate-iam-oidc-provider --cluster ${cluster_name} --region ${cluster_region} --approve
+    if [[ $? -ge 1 ]]; then
+        eksctl utils associate-iam-oidc-provider --name ${cluster_name} --region ${cluster_region} --approve
```
gautamkmr: Is this the issue with eksctl that the change the parameter name ?

scripts/generate_iam_role.sh
```diff
@@ -0,0 +1,61 @@
...
+# A function that generates an IAM role for the given account, cluster, namespace, region
+function create_namespaced_iam_role {
+    # Check if role already exists
+    aws iam get-role --role-name ${ROLE_NAME}
+    if [[ $? -eq 0 ]]; then
+        echo "A role for this cluster and namespace already exists in this account, assuming sagemaker access and proceeding."
+    else
+        echo "IAM Role does not exist, creating a new Role for the cluster"
+        aws iam create-role --role-name ${ROLE_NAME} --assume-role-policy-document file://${trustfile} --output=text
+        aws iam attach-role-policy --role-name ${ROLE_NAME}  --policy-arn arn:aws:iam::aws:policy/AmazonSageMakerFullAccess
```
gautamkmr: Attaching the policy should exist outside the "if" condition. what if role exists but it does not have policy with "SagemakerFullAccess" ?

tests/codebuild/run_integtest.sh
```diff
@@ -78,40 +103,101 @@ else
...
-        popd
-
-        echo "Waiting for controller pod to be Ready"
-        # Wait to increase chance that pod is ready
-        # TODO: Should upgrade kubectl to version that supports `kubectl wait pods --all`
-        sleep 60
-    fi 
+    pushd sagemaker-k8s-operator-install-scripts
+        # Since OPERATOR_AWS_SECRET_ACCESS_KEY and OPERATOR_AWS_ACCESS_KEY_ID defined in build spec, we will not create new user
+        ./setup_awscreds
```
gautamkmr: I am under impression that pod can assume role, if so why do we still need this?

tests/codebuild/run_integtest.sh
```diff
@@ -78,40 +103,101 @@ else
...
+        kustomize build config/crd | kubectl apply -f -
+        generate_namespace_operator_installer ${crd_namespace}
+        kubectl apply -f temp_file.yaml
+    popd
+    echo "Waiting for controller pod to be Ready"
+    # Wait to increase chance that pod is ready
+    # TODO: Should upgrade kubectl to version that supports `kubectl wait pods --all`
+    sleep 60
+    echo "Check that the manager pod is running"
+    kubectl get pods --all-namespaces | grep sagemaker
```
gautamkmr: Are we really checking anything here?

gautamkmr: probably you might want to run 
```
if [ $? eq 0] 
 success
else 
 fail
```

scripts/generate_iam_role.sh
RedbackThomson: We don't really need to support the older `eksctl` versions since the images in which we run integration tests pull the latest versions.

scripts/generate_iam_role.sh
```diff
@@ -0,0 +1,61 @@
...
+    fi
+}
+
+# Remove the generated trust file
+function cleanup {
```
RedbackThomson: I think we define a `cleanup` function in other bash scripts. Can we rename this so we don't accidentally override other functions.

scripts/generate_iam_role.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# A function that generates an IAM role for the given account, cluster, namespace, region
+function create_namespaced_iam_role {
+    # Check if role already exists
+    aws iam get-role --role-name ${ROLE_NAME}
+    if [[ $? -eq 0 ]]; then
+        echo "A role for this cluster and namespace already exists in this account, assuming sagemaker access and proceeding."
+    else
+        echo "IAM Role does not exist, creating a new Role for the cluster"
+        aws iam create-role --role-name ${ROLE_NAME} --assume-role-policy-document file://${trustfile} --output=text
```
RedbackThomson: `${trustfile}` is used both inside this function locally as well as later on globally. Perhaps we could pass it on as a parameter, instead?

tests/codebuild/common.sh
```diff
@@ -46,9 +55,10 @@ function wait_for_crd_status()
 # Cleans up all resources created during tests.
```
RedbackThomson: Can we please add the required namespace parameter in the documentation here?

tests/codebuild/create_tests.sh
```diff
@@ -4,28 +4,35 @@ source common.sh
 source inject_tests.sh
 
 # Applies each of the resources needed for the canary tests.
+# Parameter:
+#    $1: Namespace of the CRD
 function run_canary_tests
 {
-  echo "Running canary tests"
+  local crd_namespace="$1"
```
RedbackThomson: We don't want to run ALL the integration and canary tests on both cluster-scoped and namespace-scoped installations? That would effectively double our test time. Perhaps we just need to smoke test the namespacing functionality.

RedbackThomson: These smoke tests would cover all the tests cases we showed off in the demo.

tests/codebuild/run_integtest.sh
```diff
@@ -78,40 +103,101 @@ else
     readonly cluster_region="$(echo "${cluster_info}" | awk '{print $2}')"
 fi
 
-
+echo "SETUP FOR INTEGRATION TESTS"
```
RedbackThomson: Debugging output?

tests/codebuild/run_all_sample_canary_tests.sh
```diff
@@ -6,10 +6,10 @@ source delete_tests.sh
...
-run_smlogs_canary_tests
-delete_all_resources # Delete all existing resources to re-use metadata names
-run_delete_canary_tests
+run_canary_tests "default"
+run_feature_canary_tests "default"
+verify_canary_tests "default"
+verify_feature_canary_tests "default"
+run_smlogs_canary_tests "default"
+delete_all_resources "default" # Delete all existing resources to re-use metadata names
+run_delete_canary_tests "default"
```
RedbackThomson: This seems like it is redundant if we could otherwise just set a global variable (overrided by an env variable) and each of the functions would assume this as the default namespace, instead.

tests/codebuild/run_integtest.sh
```diff
@@ -1,33 +1,58 @@
...
 
-source tests/codebuild/common.sh
-
 # TODOs
 # 1. Add validation for each steps and abort the test if steps fails
 # Build environment `Docker image` has all prerequisite setup and credentials are being passed using AWS system manager
+# crd_namespace is currently hardcoded here for cleanup. Should be an env variable set in the pipeline and .env file
+
+source tests/codebuild/common.sh
+crd_namespace="test-namespace"
```
RedbackThomson: Can we make this overridable using an env variable, even if we don't actually override it in the pipeline. 

`crd_namespace=${NAMESPACE_OVERRIDE:-"test-namespace"}`

tests/codebuild/run_integtest.sh
```diff
@@ -78,40 +103,101 @@ else
...
+    pushd sagemaker-k8s-operator-install-scripts
+        # Since OPERATOR_AWS_SECRET_ACCESS_KEY and OPERATOR_AWS_ACCESS_KEY_ID defined in build spec, we will not create new user
+        ./setup_awscreds
+        path_to_installer=$(pwd)
+    popd
 popd
 
-# Run the integration test file
-cd tests/codebuild/ && ./run_all_sample_test.sh
+echo "RUN INTEGRATION TESTS FOR THE CLUSTER SCOPED OPERATOR DEPLOYMENT"
```
RedbackThomson: Debugging output?

tests/codebuild/run_integtest.sh
```diff
@@ -1,33 +1,58 @@
...
+        kubectl -n "${crd_namespace}" logs "$(kubectl get pods -n "${crd_namespace}" | grep sagemaker-k8s-operator-controller-manager | awk '{print $1}')" manager
     fi
+}
+
+# A function that cleans up Jobs, CRDs and Operator in the default namespace
+function cleanup_default_namespace {
+    set +e
+    get_manager_logs
+    delete_all_resources "default"
+    kustomize build $path_to_installer/config/default | kubectl delete -f -
```
RedbackThomson: ++ $path_to_installer

nit: Wrap all variables in double quotes: `"${path_to_installer}/config/default"`

tests/codebuild/run_integtest.sh
```diff
@@ -78,40 +103,101 @@ else
...
+pushd tests/codebuild
+    ./run_all_sample_test.sh "default"
+popd
 
 echo "Skipping private link test"
 #cd private-link-test && ./run_private_link_integration_test "${cluster_name}" "us-west-2"
+
+echo "RUN INTEGRATION TESTS FOR THE NAMESPACED OPERATOR DEPLOYMENT"
+# HELPER FUNCTIONS
+function generate_iam_role_name {
```
RedbackThomson: Please document these helper functions

tests/codebuild/run_integtest.sh
RedbackThomson: This is a debugging output more than anything. The `sleep 60` ensures it's all up and running.

tests/codebuild/run_integtest.sh
```diff
@@ -78,40 +103,101 @@ else
...
+}
+
+function generate_namespace_operator_installer {
+    local crd_namespace="$1"
+    local aws_account=$(aws sts get-caller-identity --query Account --output text)
+
+    kustomize build config/installers/rolebasedcreds/namespaced > temp_file.yaml
+    sed -i "s/PLACEHOLDER-NAMESPACE/$crd_namespace/g" temp_file.yaml
+    sed -i "s/123456789012/$aws_account/g" temp_file.yaml
+    sed -i "s/DELETE_ME/$role_name/g" temp_file.yaml
```
RedbackThomson: This is risky because if someone changes the placeholders, we break tests. Perhaps we could overlay values when we build the Kustomize targets on line 176. This probably needs more investigation - can be another task.

RedbackThomson: Probably can just add a TODO here to look into that?

scripts/generate_iam_role.sh
mbaijal: but it might fail in someone's local setup, like mine :)

scripts/generate_iam_role.sh
mbaijal: I considered this but decided against it. If I have not created a role i should not modify it IMO. On the other hand if i am reusing a namespace i have most likely already given it the required permissions. Open to changing it.

tests/codebuild/run_integtest.sh
mbaijal: because i have not modified the cluster scope operator to use the IAM role, i think that can be done as a separate task and increases the scope of this one even further.

tests/codebuild/run_integtest.sh
mbaijal: This helped me debug a couple of times and hence decided to leave it in there. the sleep guarantees sufficient time if all is well, the log confirms it.

tests/codebuild/create_tests.sh
mbaijal: Thats a good point. Just the run_canary_tests then? @gautamkmr, do you agree?

tests/codebuild/common.sh
mbaijal: Done

scripts/generate_iam_role.sh
RedbackThomson: I thought you were running this from inside the integration test container? That has a reproducible environment that we would recommend contributors to use.

tests/codebuild/create_tests.sh
RedbackThomson: Well even the canary tests don't check that the namespace scoping is correct. I think it would be nice to see something where we put a trainingjob in another namespace and make sure it never gains a status.

scripts/generate_iam_role.sh
mbaijal: done

scripts/generate_iam_role.sh
mbaijal: right, done.

tests/codebuild/run_integtest.sh
mbaijal: makes it easier to parse the logs but can remove or switch to lowercase if it's an eyesore.

tests/codebuild/run_integtest.sh
mbaijal: done

scripts/generate_iam_role.sh
mbaijal: but the script itself can be used standalone as well right.

RedbackThomson: It was never developed with that as a consideration

tests/codebuild/run_integtest.sh
gautamkmr: I am confuse, if operator has both `IAM-Role` and `Credentials` which one it will use ? So either comment this or we need to make necessary change. 

I am saying because there could be bug in `IAM Role` implementation which we might miss.

tests/codebuild/run_integtest.sh
gautamkmr: Probably change the log message in line 168, thats a bit misleading.

tests/codebuild/create_tests.sh
gautamkmr: Two things 
1. Canary tests should be just subset of tests, which can be representative. For e.g. we don't need to run FSx, EFS test for training job.
2. I think we should have a test for testing this namespace scoped functionality, as Nick mentioned. You don't need to put this test in canary but should be part of integ test

tests/codebuild/run_integtest.sh
mbaijal: The cluster scope operators are using credentials and the namespaced scope are using IAM role. 
migrating the clusterscoped operator to use the IAM role is a separate task and at that point we can get rid of credentials completely. 
Right now I can find a way to clean up the credentials before we start the namespace deployment tests. However I assume this would require some more refactoring to clean up properly.

tests/codebuild/run_integtest.sh
mbaijal: done

tests/codebuild/create_tests.sh
mbaijal: I reduced the scope of namespaced tests to one of each job type as in this function `run_canary_tests`. I can reduce this further to a single training job if you want. 

I also added a fail test as Nick mentioned

RedbackThomson: I think that new scope sounds perfect.

tests/codebuild/feature_tests.sh
```diff
@@ -14,10 +14,32 @@ function run_feature_canary_tests
...
+}
+
+# Creates a training job in the specified namespace
+# For this test, the job is created in a namespace that does not have the operator. 
+function run_feature_namespaced_tests
+{
+  echo "Running feature namespaced tests"
+  local crd_namespace="$1"
+  run_feature_canary_tests
+  run_test "${crd_namespace}" testfiles/xgboost-mnist-trainingjob.yaml
```
mbaijal: Verified the logic for this test by calling this method with the incorrect (default) and correct namespace to check that the test fails if job is created. 
```
agent_1  | trainingjob.sagemaker.aws.amazon.com/xgboost-mnist created
agent_1  | Verifying namespace deployment test
agent_1  | [SUCCESS] TrainingJob deployed to default namespace does not have a sagemaker job name
agent_1  | [FAILED] TrainingJob deployed to default namespace was created
```

tests/codebuild/run_all_sample_namespace_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+source create_tests.sh
+source delete_tests.sh
+source inject_tests.sh
+source feature_tests.sh
+
+crd_namespace="${1}"
+
+run_canary_tests "${crd_namespace}"
+verify_canary_tests "${crd_namespace}"
+run_feature_namespaced_tests "default" # Deploy job to namespace without operator.
```
RedbackThomson: Just in case we, for whatever reason, mess up the uninstallation of the cluster scoped operator, we should name this something that we know won't work. Perhaps replace `"default"` with another random namespace like `"empty-namespace"`.

tests/codebuild/run_integtest.sh
RedbackThomson: Just for consistency, I think it would be good to make it sentence cased. Could even add a square bracket header before it like `"[SECTION] Running integration tests for cluster scoped operator deployment"`.

tests/codebuild/feature_tests.sh
```diff
@@ -14,10 +14,32 @@ function run_feature_canary_tests
...
+  run_feature_canary_tests
+  run_test "${crd_namespace}" testfiles/xgboost-mnist-trainingjob.yaml  
+}
+
+# Verifies that the job created in an incorrect namespace does not gain a status or sagemaker name.
+function verify_feature_namespaced_tests
+{
+  echo "Verifying namespace deployment test"
+  local crd_namespace="$1"
+  if ! verify_trainingjob_has_no_sagemaker_name "$crd_namespace" TrainingJob "xgboost-mnist"; then
```
RedbackThomson: nit: Accept training job name as input - this is the standard we have been using.

tests/codebuild/feature_tests.sh
```diff
@@ -76,17 +101,38 @@ function verify_resource_has_additional
...
+{
+  local crd_namespace="$1"
+  local crd_type="$2"
+  local crd_instance="$3"
+  local sagemaker_training_job_name="$(kubectl get -n "$crd_namespace" "$crd_type" "$crd_instance" -o json | jq -r '.status.sageMakerTrainingJobName')"
+
+  if [ "$sagemaker_training_job_name" == null ]; then
+    echo "[SUCCESS] $crd_type deployed to $crd_namespace namespace does not have a sagemaker job name"
+    return 1
+  fi
```
RedbackThomson: I think we could also test here that it has no status. This proves the reconciler has not even considered the resource, let alone started a training job for it.

tests/codebuild/feature_tests.sh
gautamkmr: Why can't we just use `run_test` why do we need this extra function ?
Also I think you should add separate `test yaml` so that k8s name for the job are different.

tests/codebuild/feature_tests.sh
gautamkmr: Agree the status in this case should be empty, we should definitely have that check.

tests/codebuild/run_all_sample_canary_tests.sh
gautamkmr: Umm I would not be worry so much on this. That might break readability but good suggestion though,

tests/codebuild/run_integtest.sh
gautamkmr: but in namespace scope deployment do we use this code ? I think we do.

tests/codebuild/run_integtest.sh
mbaijal: Sounds good. Done.

tests/codebuild/run_integtest.sh
mbaijal: No. I commented out the cluster scoped section including the ` ./setup_awscreds` and deleted the existing user. The namespaced test ran just fine.

tests/codebuild/feature_tests.sh
mbaijal: added test yaml.

mbaijal: What do you mean by using run_test? I could have used the `wait_for_crd_status` but thats not exactly what we want here.

tests/codebuild/run_all_sample_namespace_tests.sh
mbaijal: done

tests/codebuild/feature_tests.sh
mbaijal: Done. 
```
agent_1  | Running feature namespaced tests
agent_1  | Error from server (AlreadyExists): namespaces "noop-namespace" already exists
agent_1  | Running feature canary tests
agent_1  | trainingjob.sagemaker.aws.amazon.com/xgboost-mnist-namespaced created
agent_1  | Verifying namespace deployment test
agent_1  | [SUCCESS] TrainingJob deployed to noop-namespace namespace does not have a status
agent_1  | [SUCCESS] TrainingJob deployed to noop-namespace namespace does not have a sagemaker job name
agent_1  | trainingjob.sagemaker.aws.amazon.com "xgboost-mnist-namespaced" deleted
agent_1  | Running feature namespaced tests
agent_1  | Error from server (AlreadyExists): namespaces "test-namespace" already exists
agent_1  | Running feature canary tests
agent_1  | trainingjob.sagemaker.aws.amazon.com/xgboost-mnist-namespaced created
agent_1  | Verifying namespace deployment test
agent_1  | [FAILED] TrainingJob deployed to test-namespace namespace was created and has a job status
```

tests/codebuild/feature_tests.sh
```diff
@@ -11,13 +11,50 @@ function run_feature_canary_tests
...
+  # Create the noop-namespace if it does not exist
+  kubectl create namespace "${crd_namespace}"
+  run_feature_canary_tests
+  run_test "${crd_namespace}" testfiles/xgboost-mnist-trainingjob-namespaced.yaml  
+}
+
+# Verifies that the job created in an incorrect namespace does not gain a status or sagemaker name.
+# Parameter:
+#    $1: CRD namespace
+function verify_feature_namespaced_tests
```
RedbackThomson: nit: Function name doesn't adequately describe what the function does.

tests/codebuild/feature_tests.sh
```diff
@@ -76,17 +118,57 @@ function verify_resource_has_additional
...
   done
+}
+
+# This function verifies that a given training job does not have a status.
+# Returns 1 if job creation fails.
+# Parameter:
+#    $1: CRD namespace
+#    $2: CRD type
+#    $3: Instance of CRD
+function verify_trainingjob_has_no_status
```
RedbackThomson: nit: This function accepts any crd_type not just trainingjob.

tests/codebuild/feature_tests.sh
```diff
@@ -76,17 +118,57 @@ function verify_resource_has_additional
...
+  fi
+}
+
+# This function verifies that a given training job does not have a sagemaker job name. 
+# Returns 1 if job creation fails.
+# Parameter:
+#    $1: CRD namespace
+#    $2: CRD type
+#    $3: Instance of CRD
+function verify_trainingjob_has_no_sagemaker_name
```
RedbackThomson: nit: This function accepts any crd_type not just trainingjob. However since we use `.status.sageMakerTrainingJobName` it should probably only accept trainingjobs.

tests/codebuild/run_integtest.sh
```diff
@@ -1,33 +1,58 @@
...
-function cleanup {
-    # We want to run every command in this function, even if some fail.
-    set +e
-
+# A function to print the operator logs for a given namespace
+# Parameter:
+#    $1: Namespace of the operator
+function get_manager_logs {
+    local crd_namespace="${1-sagemaker-k8s-operator-system}"
     if [ "${PRINT_DEBUG}" != "false" ]; then
```
RedbackThomson: nit: Invert logic - exit if "${PRINT_DEBUG}" == "false"

tests/codebuild/run_integtest.sh
gautamkmr: Assuming that after this change the installation will be namespace scoped which will use IAM role? 

Also do create a task in background to add support for cluster scope installation with IAM role.

mbaijal: Not sure what you mean. The local test i ran without setup_awscreds was for namespaced installation and tests. 
The integ tests do the folllowing - 
1. setup awscreds
2. Install clusterscoped operator
3. run cluster scoped tests
4. delete cluster scoped operator
5. create namespaced OIDC based IAM role
6. use #5 to deploy namespaced operator
7. Run namespaced tests. 

What i meant above was that i can run steps 5, 6, 7 without needing the aws credentials.

tests/codebuild/feature_tests.sh
mbaijal: Fixed, but open to a better suggestion.

tests/codebuild/feature_tests.sh
mbaijal: Fixed

tests/codebuild/feature_tests.sh
mbaijal: Fixed

tests/codebuild/run_integtest.sh
mbaijal: Fixing, but why is that better?

RedbackThomson: Code readability. https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html

tests/codebuild/run_integtest.sh
```diff
@@ -92,7 +93,7 @@ if [ "${need_setup_cluster}" == "true" ]; then
     readonly cluster_region="us-east-1"
 
     # By default eksctl picks random AZ, which time to time leads to capacity issue.
-    eksctl create cluster "${cluster_name}" --nodes 2 --node-type=c5.4xlarge --timeout=40m --region "${cluster_region}" --zones us-east-1a,us-east-1b,us-east-1c --auto-kubeconfig --version=1.14 --fargate 
+    eksctl create cluster "${cluster_name}" --nodes 1 --node-type=c5.xlarge --timeout=40m --region "${cluster_region}" --zones us-east-1a,us-east-1b,us-east-1c --auto-kubeconfig --version=1.14 --fargate
```
RedbackThomson, line 96: nit: Remove `--nodes` and `--node-type` from this statement - this is a Fargate cluster so they should never have been there in the first place (oops, my bad)

mbaijal: Yep, Gautam and I discussed this yesterday but was unsure if it was intentional

PR #109, opened 2020-03-30
Helm chart for namespace deployment

mbaijal:
Add Helm Charts for Namespace deployment. 
Added debugger to the helm chart. 

Tests:
Ran in a new namespace and ensured end-to-end debugger job was successful with `NoIssuesFound`

Todo:
1. Combine the cluster scope and namespaced charts.  
2. Add a Readme/update the docs

Fixes #

### All Submissions:

### DIFF
--- a/hack/charts/namespaced/crd_chart/Chart.yaml
+++ b/hack/charts/namespaced/crd_chart/Chart.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+name: amazon-sagemaker-operator-for-k8s
+version: 1.0.0
+appVersion: 1.0.1
+description: A Helm chart for deploying the Amazon SageMaker Operator CRDs for Kubernetes using EKS IAM roles for service accounts.
+maintainers:
+  - name: Gautam Kumar
+    email: gauta@amazon.com
+  - name: Cade Daniel
+    email: cdnamz@amazon.com
+  - name: Nicholas Thomson
+    email: nithomso@amazon.com
+  - name: Meghna Baijal
+    email: mbaijal@amazon.com
+keywords:
+  - amazon
+  - aws
+  - sagemaker
+  - eks
--- a/hack/charts/namespaced/crd_chart/Chart.yaml
+++ b/hack/charts/namespaced/crd_chart/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-name: amazon-sagemaker-operator-for-k8s
+name: amazon-sagemaker-operator-for-k8s-install-crds
 version: 1.0.0
 appVersion: 1.0.1
 description: A Helm chart for deploying the Amazon SageMaker Operator CRDs for Kubernetes using EKS IAM roles for service accounts.
--- a/hack/charts/namespaced/operator_chart/templates/NOTES.txt
+++ b/hack/charts/namespaced/operator_chart/templates/NOTES.txt
@@ -1,0 +1,10 @@
+Amazon SageMaker Operator for Kubernetes installed!
+
+Your release is named {{ .Release.Name }}.
+
+To learn more about the release, try:
+
+  $ helm status {{ .Release.Name }}
+  $ helm get {{ .Release.Name }}
+
+For more information about the operator: https://sagemaker.readthedocs.io/en/stable/amazon_sagemaker_operators_for_kubernetes.html
--- a/hack/charts/namespaced/operator_chart/templates/operator-installer.yaml
+++ b/hack/charts/namespaced/operator_chart/templates/operator-installer.yaml
@@ -1,0 +1,317 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: {{ .Release.Namespace | default "" | quote }}
+  ---
+kind: ServiceAccount
+metadata:
+  annotations:
+    eks.amazonaws.com/role-arn: {{ .Values.roleArn | quote }}
+  name: sagemaker-k8s-operator-default
+  namespace: {{ .Release.Namespace | default "" | quote }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: sagemaker-k8s-operator-leader-election-role
+  namespace: {{ .Release.Namespace | default "" | quote }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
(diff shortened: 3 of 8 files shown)


hack/charts/namespaced/operator_chart/templates/operator-installer.yaml
```diff
@@ -0,0 +1,317 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: {{ .Release.Namespace | default "" | quote }}
```
RedbackThomson: Here we should not offer a default namespace for any of these fields, as we require a namespace into which we should apply the resources.

hack/charts/namespaced/operator_chart/values.yaml
```diff
@@ -0,0 +1,5 @@
+roleArn: arn:aws:iam::123456789012:role/DELETE_ME
+
```
RedbackThomson: Can we add an additional option to be able to set which namespace the controller watches, in this values section? For instance if I wanted to install the operator into `my_secret_namespace` but make it watch `default`, there is no way to do that.

mbaijal:
> Is there an easy way to re-use a lot of the values we already have in the other charts? I worry about the amount of copy-paste that will need to happen when we update the CRDs. Lots of room for error in these processes.

Agreed. This problem should be solved once we combine the cluster scope and namespace installer in the next iteration. 
I think, ee should also generate these helm charts from the kuztomize target (or vise versa) to ensure these two are in sync as well.

hack/charts/namespaced/operator_chart/values.yaml
mbaijal: Oh thats a miss, will do.

hack/charts/namespaced/operator_chart/templates/operator-installer.yaml
mbaijal: If a namespace is not specified, the resources are installed into the default namespace. Why is that invalid?

hack/charts/namespaced/operator_chart/values.yaml
mbaijal: Why would i want to do that?

RedbackThomson: Some people may install their operators into a namespace that is controlled by the devops or admins. The data scientists would submit their jobs into another namespace that they have control over, or that is tied to them in some way.

hack/charts/namespaced/operator_chart/templates/operator-installer.yaml
RedbackThomson: I suppose this is fine, since if a user doesn't specify a namespace when installing the helm chart `helm install operator-installer.yaml my-operator`, it would assume `.Release.Namespace` to be `default` anyway.

hack/charts/namespaced/operator_chart/values.yaml
gautamkmr: Yeah we should have an option to pass namespace as well. 
I think behavior for namespace among helm2 and helm3 are different so don't get surprise.

gautamkmr: spoke in person namespace of operator and job has to be same. Generally helm user create namespace and add resources to it. In this case we might have to have two separate helm chart.
1. cluster scope (we already have this)
2. namespace scoped (this is what we need to do this)

hack/charts/namespaced/operator_chart/templates/operator-installer.yaml
mbaijal: true, so we could have done without the default value actually. Leaving this be.

hack/charts/namespaced/operator_chart/values.yaml
RedbackThomson: Oh? Why does the namespace of the operator and the job have to be the same?

mbaijal: @RedbackThomson I think that is a valid but edge case. Even with the current implementation if someone wanted to achieve this, they could modify the templates before deploying the helm. 
On the other hand adding an extra namespace option to the values.yaml file would be confusing to the majority of the users. I say this after reading posts on helm usage ([link this one](https://github.com/helm/helm/issues/5628)) which indicate that helm2 vs helm3 already adds some confusion about which namespace helm deploys resources to. Lets keep it simple for the majority.

RedbackThomson: Sure! Totally reasonable, yep.

gautamkmr: > Oh? Why does the namespace of the operator and the job have to be the same?

don't we have operator running in namespace will pick the job only in that namespace ?

RedbackThomson: > > Oh? Why does the namespace of the operator and the job have to be the same?
> 
> don't we have operator running in namespace will pick the job only in that namespace ?

The namespace in which the operator is deployed does not need to be the one it monitors. They are separate fields that can be set independently. Whether RBAC permits that is a different question, but in theory it's entirely possible.

gautamkmr: Yeah thats definitely confusing, let user define the namespace if they want it separate or an option.

hack/charts/namespaced/crd_chart/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+name: amazon-sagemaker-operator-for-k8s
```
RedbackThomson: nit: Chart name doesn't specify that it's only CRDs

hack/charts/namespaced/operator_chart/templates/NOTES.txt
```diff
@@ -0,0 +1,10 @@
+Amazon SageMaker Operator for Kubernetes installed!
```
RedbackThomson: nit: Does not specify this is the operator

hack/charts/namespaced/crd_chart/Chart.yaml
mbaijal: Fixed

hack/charts/namespaced/operator_chart/templates/NOTES.txt
mbaijal: I think this is the cleanest way of writing it, have not changed this.

PR #114, opened 2020-04-06
Use IAM role for cluster scope operator deployment

mbaijal:
Use IAM role for cluster scope operator deployment, removed the usage of aws credentials. 

Tested branch in pipeline.

Fixes #

### All Submissions:

### DIFF
--- a/tests/codebuild/run_integtest.sh
+++ b/tests/codebuild/run_integtest.sh
@@ -34,7 +34,7 @@ function cleanup {
         kustomize build "$path_to_installer/config/default" | kubectl delete -f -
         # Delete the namespaced operator. TODO: This can be cleaner if parameterized
         kustomize build "$path_to_installer/config/crd" | kubectl delete -f -
-        rolebased_operator_installation ${crd_namespace} "config/installers/rolebasedcreds/namespaced" "${role_name}" "delete"  
+        rolebased_operator_install_or_delete ${crd_namespace} "config/installers/rolebasedcreds/namespaced" "${role_name}" "delete"  
     fi
 
     if [ "${existing_fsx}" == "false" ] && [ "$FSX_ID" != "" ]; then
@@ -106,7 +106,7 @@ popd
 #    $3: Name of the IAM Role to use
 #    $4: kubectl action - apply or delete the resources defined in the installer
 # TODO:  Investigate if it is possible to overlay values when we build the Kustomize targets instead. 
-function rolebased_operator_installation {
+function rolebased_operator_install_or_delete {
     local crd_namespace="$1"
     local kustomize_file_path="$2"
     local role="$3"
@@ -134,7 +134,7 @@ if [ "${SKIP_INSTALLATION}" == "true" ]; then
 else
     pushd sagemaker-k8s-operator/sagemaker-k8s-operator-install-scripts
         echo "Deploying the operator to the default namespace"
-        rolebased_operator_installation "${default_operator_namespace}" "config/installers/rolebasedcreds" "${default_role_name}" "apply"
+        rolebased_operator_install_or_delete "${default_operator_namespace}" "config/installers/rolebasedcreds" "${default_role_name}" "apply"
     popd
     echo "Waiting for controller pod to be Ready"
     # Wait to increase chance that pod is ready
@@ -162,7 +162,7 @@ function operator_namespace_deploy {
     # Goto directory that holds the CRD 
     pushd sagemaker-k8s-operator/sagemaker-k8s-operator-install-scripts
         kustomize build config/crd | kubectl apply -f -
-        rolebased_operator_installation ${crd_namespace} "config/installers/rolebasedcreds/namespaced" "${role_name}" "apply"
+        rolebased_operator_install_or_delete ${crd_namespace} "config/installers/rolebasedcreds/namespaced" "${role_name}" "apply"
     popd
     echo "Waiting for controller pod to be Ready"
     # Wait to increase chance that pod is ready
--- a/tests/codebuild/run_integtest.sh
+++ b/tests/codebuild/run_integtest.sh
@@ -125,6 +125,8 @@ function generate_iam_role_name {
     local cluster=$(echo ${cluster_name} | cut -d'/' -f2)
 
     role_name="${cluster}-${crd_namespace}"
+    # IAM Role name must have length less than 64
+    role_name=`echo $role_name|cut -c1-64`
 }
 
 # A helper function that generates the namespace-scoped operator installer yaml file with updated namespace and role. 
--- a/tests/codebuild/run_integtest.sh
+++ b/tests/codebuild/run_integtest.sh
@@ -6,6 +6,7 @@
 # crd_namespace is currently hardcoded here for cleanup. Should be an env variable set in the pipeline and .env file
 
 source tests/codebuild/common.sh
+default_operator_namespace="sagemaker-k8s-operator-system"
 crd_namespace=${NAMESPACE_OVERRIDE:-"test-namespace"}
 
 # Verbose trace of commands, helpful since test iteration takes a long time.
@@ -15,7 +16,7 @@ set -x
 # Parameter:
 #    $1: Namespace of the operator
 function get_manager_logs {
-    local crd_namespace="${1-sagemaker-k8s-operator-system}"
+    local crd_namespace="${1:-$default_operator_namespace}"
     if [ "${PRINT_DEBUG}" != "false" ]; then
         echo "Controller manager logs in the ${crd_namespace} namespace"
         kubectl -n "${crd_namespace}" logs "$(kubectl get pods -n "${crd_namespace}" | grep sagemaker-k8s-operator-controller-manager | awk '{print $1}')" manager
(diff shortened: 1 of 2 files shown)


tests/codebuild/run_integtest.sh
```diff
@@ -113,22 +114,55 @@ tar -xf sagemaker-k8s-operator.tar.gz
 pushd sagemaker-k8s-operator
 # Setup the PATH for smlogs
     mv smlogs-plugin/linux.amd64/kubectl-smlogs /usr/bin/kubectl-smlogs
-    pushd sagemaker-k8s-operator-install-scripts
-        # Since OPERATOR_AWS_SECRET_ACCESS_KEY and OPERATOR_AWS_ACCESS_KEY_ID defined in build spec, we will not create new user
-        ./setup_awscreds
-        path_to_installer=$(pwd)
-    popd
+    path_to_installer=$(pwd)/sagemaker-k8s-operator-install-scripts
```
RedbackThomson, line 117: I hate this tarball problem. I will try to get rid of this in my next change, though.

tests/codebuild/run_integtest.sh
```diff
@@ -6,6 +6,7 @@
 # crd_namespace is currently hardcoded here for cleanup. Should be an env variable set in the pipeline and .env file
 
 source tests/codebuild/common.sh
+default_operator_namespace="sagemaker-k8s-operator-system"
 crd_namespace=${NAMESPACE_OVERRIDE:-"test-namespace"}
```
RedbackThomson, line 10: I just realised that this variable name doesn't make much sense. The CRDs are global, so there is no "CRD namespace". I understand this was for the namespace-scope tests, so I should have caught it in the previous PR.

tests/codebuild/run_integtest.sh
```diff
@@ -113,22 +114,55 @@ tar -xf sagemaker-k8s-operator.tar.gz
...
     mv smlogs-plugin/linux.amd64/kubectl-smlogs /usr/bin/kubectl-smlogs
-    pushd sagemaker-k8s-operator-install-scripts
-        # Since OPERATOR_AWS_SECRET_ACCESS_KEY and OPERATOR_AWS_ACCESS_KEY_ID defined in build spec, we will not create new user
-        ./setup_awscreds
-        path_to_installer=$(pwd)
-    popd
+    path_to_installer=$(pwd)/sagemaker-k8s-operator-install-scripts
 popd
 
+# A helper function to generate an IAM Role name for the current cluster and sepcified namespace
```
RedbackThomson, line 120: nit: s/sepcified/specified/

tests/codebuild/run_integtest.sh
```diff
@@ -113,22 +114,55 @@ tar -xf sagemaker-k8s-operator.tar.gz
...
+# Parameter:
+#    $1: Namespace of CRD
+function generate_iam_role_name {
+    local crd_namespace="$1"
+    local cluster=$(echo ${cluster_name} | cut -d'/' -f2)
```
RedbackThomson, line 125: nit: Wrap in double quotes

tests/codebuild/run_integtest.sh
```diff
@@ -113,22 +114,55 @@ tar -xf sagemaker-k8s-operator.tar.gz
...
+    role_name=`echo $role_name|cut -c1-64`
+}
+
+# A helper function that generates the namespace-scoped operator installer yaml file with updated namespace and role. 
+# Parameter:
+#    $1: Namespace to which the op pod is installed. For cluster scope, this value is not used. 
+#    $2: Path to the kustomize source
+#    $3: Name of the IAM Role to use
+# TODO:  Investigate if it is possible to overlay values when we build the Kustomize targets instead. 
+function generate_operator_installer_for_given_role {
```
RedbackThomson, line 138: This method should take an output path as a parameter, since we reference `temp_file.yaml` elsewhere in the scripts.

tests/codebuild/run_integtest.sh
```diff
@@ -113,22 +114,55 @@ tar -xf sagemaker-k8s-operator.tar.gz
...
+    local crd_namespace="$1"
+    local file_path="$2"
+    local role="$3"
+    local aws_account=$(aws sts get-caller-identity --query Account --output text)
```
RedbackThomson, line 142: Should we make this a global variable, it seems we use it quite a few times?

tests/codebuild/run_integtest.sh
```diff
@@ -113,22 +114,55 @@ tar -xf sagemaker-k8s-operator.tar.gz
...
+    sed -i "s/PLACEHOLDER-NAMESPACE/$crd_namespace/g" temp_file.yaml
+    sed -i "s/123456789012/$aws_account/g" temp_file.yaml
+    sed -i "s/DELETE_ME/$role/g" temp_file.yaml
+}
+
+
 echo "[SECTION] Run integration tests for the cluster scoped operator deployment"
+
+generate_iam_role_name "${default_operator_namespace}"
+cd scripts && ./generate_iam_role.sh "${cluster_name}" "${default_operator_namespace}" "${role_name}" "${cluster_region}" && cd ..
```
RedbackThomson, line 154: nit: Why do we need the `cd` for this? Could we not simply use the larger path `scripts/generate_iam_role.sh`?

tests/codebuild/run_integtest.sh
mbaijal: This was a todo in my previous PR. You used the variable crd_namespace in your branch, which i picked and now its everywhere. Needs a massive cleanup.

RedbackThomson: Oh oops. I was going to do a big cleanup in my next refactor, so that's fine I can take care of it.

tests/codebuild/run_integtest.sh
```diff
@@ -27,7 +28,19 @@ function cleanup_default_namespace {
     set +e
     get_manager_logs
     delete_all_resources "default"
-    kustomize build "$path_to_installer/config/default" | kubectl delete -f -
+    generate_operator_installer_for_given_role "${default_operator_namespace}" "config/installers/rolebasedcreds" "${default_role_name}"
+    kubectl delete -f temp_file.yaml
+    rm temp_file.yaml
+}
+
+# A function that cleans up Jobs, CRDs and Operator in the default namespace
```
RedbackThomson, line 36: nit: Function does not match comment.

tests/codebuild/run_integtest.sh
```diff
@@ -27,7 +28,19 @@ function cleanup_default_namespace {
...
+# Parameter:
+#    $1: Name of the Role to be deleted
+function delete_generated_role {
+    local role_to_delete="${1}"
+    # Delete the role associated with the cluster thats being deleted
+    aws iam detach-role-policy --role-name "${role_to_delete}" --policy-arn arn:aws:iam::aws:policy/AmazonSageMakerFullAccess
```
RedbackThomson, line 42: Duplication of role policies, here. Perhaps we could store the list of policies in an array, then when we create we iterate over and apply - likewise for delete.

tests/codebuild/run_integtest.sh
```diff
@@ -45,15 +58,16 @@ function cleanup {
...
+        delete_generated_role "${default_role_name}"
     else
         echo "need_setup_cluster is not true, will remove operator without deleting cluster"
         kustomize build "$path_to_installer/config/default" | kubectl delete -f -
         # Delete the namespaced operator. TODO: This can be cleaner if parameterized
         kustomize build "$path_to_installer/config/crd" | kubectl delete -f -
-        generate_namespace_operator_installer "${crd_namespace}"
-        kubectl delete -f temp_file.yaml    
+        generate_operator_installer_for_given_role ${crd_namespace} "config/installers/rolebasedcreds/namespaced" "${role_name}"
+        kubectl delete -f temp_file.yaml
```
gautamkmr, line 69: Reading from above it doesn't quite tells me where did `temp_file.yaml` come from? So either change the function to take the file path or echo the output from that function and write it here in file in line 68.

tests/codebuild/run_integtest.sh
mbaijal: I think its cleaner to apply/delete the generated installer and delete it within the scope of the method itself instead of carrying around different filenames. Changing accordingly.

RedbackThomson: In that case, we should rename it to something more descriptive.

tests/codebuild/run_integtest.sh
mbaijal: yea, just carried forward from old code, cleaning.

tests/codebuild/run_integtest.sh
mbaijal: good point, i cleaned up this logic by limiting the use of the temp_file.yaml to the function where it is generated.

tests/codebuild/run_integtest.sh
mbaijal: Not sure i understand, the same policy is getting attached to each role.

RedbackThomson: I foresee us adding extra permissions in the future (for CloudWatch logs and EC2 metrics). I suppose this is a nit since this isn't the case at the moment.

tests/codebuild/run_integtest.sh
mbaijal: > Could we move some of the common functions out into a utilities/helper script file? It seems that there are a few IAM and generator functions that break up the flow of this file quite a bit.

Ok if i move them to the common.sh?

mbaijal: Its used only twice, I think it ok as is for now.

tests/codebuild/run_integtest.sh
```diff
@@ -45,15 +27,14 @@ function cleanup {
...
+        delete_generated_role "${role_name}"
+        delete_generated_role "${default_role_name}"
     else
         echo "need_setup_cluster is not true, will remove operator without deleting cluster"
         kustomize build "$path_to_installer/config/default" | kubectl delete -f -
         # Delete the namespaced operator. TODO: This can be cleaner if parameterized
         kustomize build "$path_to_installer/config/crd" | kubectl delete -f -
-        generate_namespace_operator_installer "${crd_namespace}"
-        kubectl delete -f temp_file.yaml    
+        rolebased_operator_installation ${crd_namespace} "config/installers/rolebasedcreds/namespaced" "${role_name}" "delete"
```
gautamkmr, line 37: Its a bit misnomer. You are calling installation but action can be delete as well. probably just change it to something different `rolebased_operator_operation`

mbaijal: Fair, updated.

PR #127, opened 2020-07-24
Integration Test for HostingDeployment Autoscaling

mbaijal:
Added an integration Test for the HostingDeploymentAutoscalingJob - 
1. In order to test multiple endpoint/variant pairs I create a second hostingdeployment job
2. I wait for both to complete and then capture the sagemaker name of the endpoint
3. I added an additional check where it uses the awscli to make sure a scalingPolicy is applied

 - Coming up - a custom policy test
 - The CRD name to be changed after these PRs are merged

No

### All Submissions:

### DIFF
--- a/tests/codebuild/create_tests.sh
+++ b/tests/codebuild/create_tests.sh
@@ -23,6 +23,7 @@ function run_canary_tests
   yq w -i testfiles/xgboost-mnist-batchtransform.yaml "spec.modelName" "$(get_sagemaker_model_from_k8s_model "${crd_namespace}" xgboost-model)"
   run_test "${crd_namespace}" testfiles/xgboost-mnist-batchtransform.yaml 
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
+  run_hap_test "${crd_namespace}" xgboost-hosting testfiles/xgboost-hostingautoscaling.yaml
   run_test "${crd_namespace}" testfiles/xgboost-mnist-trainingjob-debugger.yaml
 }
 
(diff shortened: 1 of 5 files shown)


tests/codebuild/create_tests.sh
```diff
@@ -85,6 +87,66 @@ function verify_integration_tests
...
+
+  # Get the sagemaker names of the two hostingdeployment endpoints and replace in hap yaml
+  local endpoint_name_1="$(kubectl get -n "$target_namespace" "$hostingdeployment_type" "$hosting_deployment_1" -o=custom-columns=SAGEMAKER_ENDPOINT-NAME:.status.endpointArn | tail -1 | cut -d'/' -f2)"
+  local endpoint_name_2="$(kubectl get -n "$target_namespace" "$hostingdeployment_type" "$hosting_deployment_2" -o=custom-columns=SAGEMAKER_ENDPOINT-NAME:.status.endpointArn | tail -1 | cut -d'/' -f2)"
+
+  sed -i "s/PLACEHOLDER-ENDPOINT-1/$endpoint_name_1/g" "$file_name"
+  sed -i "s/PLACEHOLDER-ENDPOINT-2/$endpoint_name_2/g" "$file_name"
+  run_test "$target_namespace" "$file_name"
+
+  kubectl describe hostingdeploymentautoscalingjob
```
RedbackThomson, line 120: Describe our specific HAP?

tests/codebuild/create_tests.sh
```diff
@@ -85,6 +87,66 @@ function verify_integration_tests
...
+function verify_hap_test()
+{
+  local crd_namespace="$1"
+  local crd_type="$2"
+  local crd_instance="$3"
+  local timeout="$4"
+  local desired_status="$5"
+
+  verify_test "${crd_namespace}" "${crd_type}" "${crd_instance}" "${timeout}" "${desired_status}"
+  scaling_policy="$(aws application-autoscaling describe-scaling-policies --region us-west-2 --service-namespace sagemaker | jq .ScalingPolicies[0])"
```
RedbackThomson, line 140: We don't need to rely on jq if we're using an awscli command. `--output=json --query="ScalingPolicies[0]"` will achieve the same effect.

tests/codebuild/create_tests.sh
```diff
@@ -85,6 +87,66 @@ function verify_integration_tests
...
+
+  if [ "${scaling_policy}"=="null" ]; then
```
RedbackThomson, line 142: If using the above method, just double check the default, here. It might just be empty string.

tests/codebuild/create_tests.sh
mbaijal: Debugding, removed

mbaijal:
Merged this into the ep-as branch, will update review comments on that branch and resolve here.

tests/codebuild/create_tests.sh
mbaijal: good tip but have not changed at the moment. It took me a while to get this working though so this would have helped!

tests/codebuild/create_tests.sh
mbaijal: yea, I ran it manually several times to check the error condition.

PR #128, opened 2020-07-24
Canary tests for China region

akartsky:
Canary tests for China region 

This uses codebuild job to run the Dockerfile.canary.china container 

Notable things :
- This runs a subset of tests being run in US 
- Installer_china.yaml generation using Kustomize requires a hacky solution (no better way to do it)
- Codebuild uses docker caching (check build-spec)
- code is region specific can't be deployed to US region directly


release/rolebased/installer_china.yaml
```diff
@@ -0,0 +1,4110 @@
+apiVersion: v1
```
RedbackThomson, line 1: How was this file produced? We have a Make target to produce our installer YAMLs using Kustomize that should be extended (and modified) to also do these.

tests/codebuild/run_canarytest_china.sh
```diff
@@ -0,0 +1,97 @@
...
+    ]
+  }
+  ' > ./trust.json
+
+  aws --region $CLUSTER_REGION iam create-role --role-name $OIDC_ROLE_NAME --assume-role-policy-document file://trust.json
+  aws --region $CLUSTER_REGION iam attach-role-policy --role-name $OIDC_ROLE_NAME --policy-arn arn:aws-cn:iam::aws:policy/AmazonSageMakerFullAccess
+  OIDC_ROLE_ARN=$(aws --region $CLUSTER_REGION iam get-role --role-name $OIDC_ROLE_NAME --output text --query 'Role.Arn')
+
+  wget --retry-connrefused --waitretry=30 --read-timeout=20 --timeout=15 -t 30 \
+       -O installer_china.yaml https://raw.githubusercontent.com/akartsky/amazon-sagemaker-operator-for-k8s/china_test/release/rolebased/installer_china.yaml
```
RedbackThomson, line 61: This needs to be supported on the `master` branch. I understand this is for testing, now, but must be changed before merge.

tests/codebuild/run_canarytest_china.sh
```diff
@@ -0,0 +1,97 @@
...
+
+  OIDC_URL=$(aws eks describe-cluster --region $CLUSTER_REGION --name $CLUSTER_NAME --query "cluster.identity.oidc.issuer" --output text | cut -c9-)
+
+  printf '{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Effect": "Allow",
+        "Principal": {
+          "Federated": "arn:aws-cn:iam::'$AWS_ACC_NUM':oidc-provider/'$OIDC_URL'"
```
RedbackThomson, line 42: I feel that the `aws-cn` prefix could easily be made into a variable, here, which should allow for easier integration back into the US system.

tests/codebuild/run_canarytest_china.sh
```diff
@@ -0,0 +1,97 @@
...
+OIDC_ROLE_NAME=pod-role-$DEPLOYMENT_NAME
+AWS_ACC_NUM=$(aws sts get-caller-identity --region $CLUSTER_REGION   --query Account --output text)
+
+
+function create_eks_cluster() {
+  eksctl create cluster --name $CLUSTER_NAME --region $CLUSTER_REGION --auto-kubeconfig --timeout=30m --managed --node-type=c5.xlarge --nodes=1
+}
+
+function install_k8s_operators() {
+  echo installing SageMaker operators
```
RedbackThomson, line 30: nit: Quote around echo statements
nit: Capitalise "i" in "installing"

tests/codebuild/testfiles/xgboost-mnist-hpo-china.yaml
```diff
@@ -0,0 +1,109 @@
+apiVersion: sagemaker.aws.amazon.com/v1
+kind: HyperparameterTuningJob
+metadata:
+  name: xgboost-mnist-hpo
```
RedbackThomson, line 4: nit: Add `-china` to name.

tests/codebuild/testfiles/xgboost-mnist-batchtransform-china.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: sagemaker.aws.amazon.com/v1
+kind: BatchTransformJob
+metadata:
+  name: xgboost-batch
```
RedbackThomson, line 4: nit: Add `-china` to name.

tests/images/Dockerfile.canary.china
```diff
@@ -0,0 +1,35 @@
+FROM ubuntu:18.04
```
RedbackThomson, line 1: I understand this is a simplified version of the other container, but I don't think we should be duplicating code. It would be preferable to extend the other image (or simply use the other image) if all we are doing is changing the entrypoint? 

This can be done by either:
1. Building the China image with the US image as the base
2. Running the US image with a different entrypoint when doing `docker run`.

release/rolebased/installer_china.yaml
akartsky: done

Makefile
```diff
@@ -101,9 +111,15 @@ else
 CONTROLLER_GEN=$(shell which controller-gen)
 endif
 
-create-installers: set-image
+create-installers-china: modify-base-kustomize-china
+	kustomize build config/installers/rolebasedcreds > $(INSTALLER_PATH)/rolebased/installer_china.yaml
+	kustomize build config/installers/rolebasedcreds/namespaced > $(INSTALLER_PATH)/rolebased/namespaced/operator_china.yaml
```
RedbackThomson, line 116: I think we should have a separate folder for the Chinese releases so people don't accidentally try to deploy them elsewhere. Perhaps just a subfolder.

config/installers/rolebasedcreds/namespaced/kustomization.yaml
```diff
@@ -57,6 +57,18 @@ patchesJson6902:
     - op: replace
       path: "/subjects/0/name"
       value: sagemaker-k8s-operator-default
+# Hacky solution so that its possible to add --namespace for both China and US region with this one file
+# change the position of pods so that manager pod is first pod in both US and China
```
RedbackThomson, line 61: Sorry I don't quite follow. Why is this necessary?

config/base/kustomization.yaml
```diff
@@ -3,14 +3,12 @@ kind: Kustomization
...
 
-patchesStrategicMerge:
-- manager_auth_proxy_patch.yaml
 # Add other common files from base dir if you need to
 
 # This is set by `make docker-build` but not when you build helm chart
 images:
 - name: controller
-  newName: 957583890962.dkr.ecr.us-east-1.amazonaws.com/amazon-sagemaker-operator-for-k8s
+  newName: 099223943020.dkr.ecr.cn-northwest-1.amazonaws.com.cn/amazon-sagemaker-operator-for-k8s
```
mbaijal, line 11: We are replacing the existing ?

tests/codebuild/run_canarytest_china.sh
mbaijal: Agreed, this is a lot of code duplication and updates will need to be made both places. 
The test yaml files should also not be copied, inject/replace the region into them.

akartsky: Yep 

Lot of other stuff need to be changed to support REGION variable 
will take this up as post release task

tests/images/Dockerfile.canary.china
akartsky: Yep 
once we have same infra in US and China region we can just use 1 Dockerfile for both 

-------
Currently there are some issues like
 * sagemaker-k8s-operator.tar.gz doesn't exist for China canary 
 * Kustomize fails to download

config/base/kustomization.yaml
akartsky: Ignore this file change

Makefile
akartsky: done

config/installers/rolebasedcreds/namespaced/kustomization.yaml
akartsky: manager pod is at index 1 in US and index 0 in China 

path: "/spec/template/spec/containers/0/args/-"
doesn't allow us to specify last item is an array with index -1 

moving the manager pod so that it is at index 0 in both US and China

tests/codebuild/buildspec_china/canary-china.buildspec.yaml
```diff
@@ -0,0 +1,11 @@
+version: 0.2
+phases:
+  build:
+    commands:
+      - cp release/rolebased/china/installer_china.yaml tests/codebuild/
+      - cd tests
+      - aws ecr get-login-password --region $CLUSTER_REGION | docker login --username AWS --password-stdin $ECR_CACHE_URI || true
```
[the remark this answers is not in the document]
akartsky: No it's the other way around 
"get-login" is deprecated 

https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login.html#description

tests/codebuild/buildspec_china/canary-china.buildspec.yaml
```diff
@@ -0,0 +1,11 @@
...
+      - docker pull ${ECR_CACHE_URI}:latest || true
+      - docker build -f images/Dockerfile.canary.china . -t ${ECR_CACHE_URI}:latest --quiet || true
```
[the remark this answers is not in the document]
akartsky: If the build fails it will use the cached image 
which might use older versions of our dependencies but that should be okay 
(btw the build fails around 20% of the time as far as i have observed)

i'm injecting our latest code during docker run using "--mount"
so the code will be always the latest one from github even if we use cached image

tests/codebuild/buildspec_china/canary-china.buildspec.yaml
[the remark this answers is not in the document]
RedbackThomson: Oh wow I'm completely wrong. Thanks for letting me know :)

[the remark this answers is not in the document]
akartsky: 😄

PR #137, opened 2020-08-25
fix: create installer should happen after generating CRDs

surajkota:
bug fix in generating CRD

No

### Testing:
Make install and manifests work fine and crd get installed on cluster as expected.
Synced up with @RedbackThomson to make sure other make targets are not affected by this change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,10 @@ run: lint generate fmt vet
 install: manifests
 	kubectl apply -f config/crd/bases
 
+# Generate the CRDs, RBAC etc. install the CRDs onto your cluster and create installers to be keep in sync with CRDs
+# Recommended to use this as opposed to just install if you intend to check-in code
+generate_and_install: install create-installers
+
 # Build a tarball containing everything needed to install the operator onto a cluster.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ undeploy: manifests
 
 # Generate manifests e.g. CRD, RBAC etc.
 # Requires golang development setup for controller-gen.
-manifests: controller-gen create-installers
+manifests: controller-gen
```
mbaijal, line 58: But this means we are not generating installers in `make` and `make install` at all ? The idea was to ensure these changes are checked in with PRs.

RedbackThomson: Perhaps we could move what's currently in `manifests` into another target called `create-crds` and then have `manifests` be `manifests: create-crds create-installers`. That way when doing `make manifests` it creates CRDs and installers (as I suppose this change was originally intended to do).

mbaijal: yep exactly!

surajkota: didnt want to change the default kubebuilder make targets, created a new target generate_and_install which does as the name says

PR #138, opened 2020-08-28
Improve China canary tests

akartsky:
Improve China canary tests 

1] Use code from S3 bucket
2] Give option to create EKS with existing VPC
3] Update wget location


codebuild/run_canary_china.yaml
```diff
@@ -2,6 +2,8 @@ version: 0.2
 phases:
   build:
     commands:
+      # make all shell scripts executable. This is required when running code copyied from S3
+      - find ./ -type f -name "*.sh" -exec chmod +x {} \;
```
RedbackThomson, line 6: nit: `\;` at the end is not necessary.

tests/codebuild/run_canarytest_china.sh
```diff
@@ -48,7 +57,26 @@ function download_installer_china(){
 }
 
 function create_eks_cluster() {
-  eksctl create cluster --name $CLUSTER_NAME --region $CLUSTER_REGION --auto-kubeconfig --timeout=30m --managed --node-type=c5.xlarge --nodes=1
+  if [[ $USE_EXISTING_SUBNET ]]; then
```
RedbackThomson, line 60: Here is a more elegant way to optionally add the VPC flags:
https://github.com/kubeflow/pipelines/blob/ada18bc6e604b9eaafa6fdad3e7a671b61606d75/components/aws/sagemaker/tests/integration_tests/scripts/run_integration_tests#L105-L107

codebuild/run_canary_china.yaml
akartsky: it is

tests/codebuild/run_canarytest_china.sh
akartsky: Sweet ! .. thanks for pointing it 
will change

codebuild/run_canary_china.yaml
RedbackThomson: `\` denotes the command continues to the next line? and then `;` denotes the end of the command? Is this a weird function of the `-exec` flag?

akartsky: ya so the syntax is like this 
`find <something> -exec <command> ;`

but if you just type `;` at it will think that it's part of the <command>
("chmod +x {}" ; does not work either)
so you need to escape it with `\;`

more info 
https://stackoverflow.com/questions/25142822/error-find-exec-no-terminating-or
https://ss64.com/bash/find.html

tests/codebuild/run_canarytest_china.sh
akartsky: done

RedbackThomson (approved):

tests/codebuild/run_canarytest_china.sh
```diff
@@ -48,7 +57,11 @@ function download_installer_china(){
 }
 
 function create_eks_cluster() {
-  eksctl create cluster --name $CLUSTER_NAME --region $CLUSTER_REGION --auto-kubeconfig --timeout=30m --managed --node-type=c5.xlarge --nodes=1
+  eksctl_args=( --managed --nodes 1 --node-type=c5.xlarge --timeout=30m --region "$CLUSTER_REGION" --auto-kubeconfig )
```
mbaijal, line 60: These configurations should not be hardcoded. If necessary, set as default with an option to override.

mbaijal (approved):

PR #139, opened 2020-09-11
Add EndpointName option to HostingDeployment

RedbackThomson:
Adds an `EndpointName` option to the HostingDeployment spec. This allows users to set a deterministic name for the SageMaker endpoint. If an endpoint exists with that name already, the operator will overwrite the corresponding endpoint config with one it generates.

N/A

### All Submissions:


controllers/hosting/hostingdeployment_controller_test.go
```diff
@@ -217,15 +217,11 @@ var _ = Describe("Reconciling a HostingDeployment that exists", func() {
...
-
-		BeforeEach(func() {
-			mockSageMakerClientBuilder.
-				AddDescribeEndpointErrorResponse(clientwrapper.DescribeEndpoint404Code, clientwrapper.DescribeEndpoint404MessagePrefix, 400, "request id")
-		})
-
 		Context("HasDeletionTimestamp", func() {
-
 			BeforeEach(func() {
+				mockSageMakerClientBuilder.
```
mbaijal, line 222: Just cleanup ?

controllers/hosting/hostingdeployment_controller_test.go
```diff
@@ -245,39 +241,84 @@ var _ = Describe("Reconciling a HostingDeployment that exists", func() {
...
-				Expect(*createdRequest.EndpointName).To(Equal(controllercommon.GetGeneratedJobName(deployment.ObjectMeta.GetUID(), deployment.ObjectMeta.GetName(), 63)))
-			})
+				It("Recognises the existing endpoint", func() {
+					ExpectStatusToBe(deployment, string(sagemaker.EndpointStatusInService))
+				})
 
-			It("Requeues after interval", func() {
-				ExpectRequeueAfterInterval(reconcileResult, reconcileError, pollDuration)
+				It("Creates new resources", func() {
+					ExpectNthSubreconcilerCallToKeepUnusedResources(modelReconciler, endpointConfigReconciler, 0)
```
mbaijal, line 270: Could you explain what we are testing in this overall Context please and also specifically line 269.

config/crd/bases/sagemaker.aws.amazon.com_hostingautoscalingpolicies.yaml
```diff
@@ -29,18 +29,24 @@ spec:
     status: {}
   validation:
     openAPIV3Schema:
-      description: HostingAutoscalingPolicy is the Schema for the HostingAutoscalingPolicy API
+      description: HostingAutoscalingPolicy is the Schema for the HostingAutoscalingPolicy
+        API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation
```
surajkota, line 36: why is this PR changing other CRDs

RedbackThomson: I had to update the `go.mod` file because it was referencing older versions that had broken upstream dependencies.

controllers/hosting/hostingdeployment_controller_test.go
RedbackThomson: Yes had to create a new context to use the deployment with the `endpointName`, so most of the removes here are just me moving code around to fit.

controllers/hosting/hostingdeployment_controller_test.go
```diff
@@ -1058,15 +1149,28 @@ func ExpectRequestToDeleteHostingDeployment(req interface{}, deployment *hosting
...
+	Expect(*deleteRequest.EndpointName).To(Equal(endpointName))
 }
 
 // Helper function to verify that the specified object is n UpdateEndpointInput and that it requests to update the HostingDeployment correctly.
 func ExpectRequestToUpdateHostingDeployment(req interface{}, deployment *hostingv1.HostingDeployment, expectedEndpointConfigName string) {
 	Expect(req).To(BeAssignableToTypeOf((*sagemaker.UpdateEndpointInput)(nil)))
 
 	updateRequest := req.(*sagemaker.UpdateEndpointInput)
-	Expect(*updateRequest.EndpointName).To(Equal(controllercommon.GetGeneratedJobName(deployment.ObjectMeta.GetUID(), deployment.ObjectMeta.GetName(), 63)))
+	var endpointName string
```
mbaijal, line 1167: Does this validate the case if the spec had no endpointname originally but is updated to have one and re-applied ?

controllers/hosting/hostingdeployment_controller_test.go
```diff
@@ -318,6 +359,19 @@ var _ = Describe("Reconciling a HostingDeployment that exists", func() {
 					})
 				})
 
+				Context("EndpointName is defined", func() {
```
mbaijal, line 362: nit: A better message might be something like "deterministic endpoint name" since that's what we have been calling it. 
I can imagine someone being a little confused by this context in a few months.

controllers/hosting/endpointconfig_reconciler.go
```diff
@@ -78,7 +78,9 @@ func (r *endpointConfigReconciler) Reconcile(ctx context.Context, desiredDeploym
 		return errors.Wrap(err, "Unable to interpret HostingDeployment endpoint config")
 	}
 
-	r.log.Info("Desired endpoint config", "desired", desiredEndpointConfig)
+	if desiredEndpointConfig != nil {
```
surajkota, line 81: if this is ever nil, the job wont be accepted by API server because some of the fields in the endpointconfig would be mandatory
 
even if it gets past that extraction would fail and that would be an error in previous block itself ?

RedbackThomson: I only added this because I was getting panics because of this `log`. This isn't anything to do with this PR, necessarily. I found this is an older PR, this is an ongoing issue.

controllers/hosting/hostingdeployment_controller_test.go
RedbackThomson: I am not exactly sure, to be honest. This test case, from the comments, ensures that the n-th subreconciler call did not delete the resources. These are the expects that the other test cases use in this scenario.

controllers/hosting/hostingdeployment_controller_test.go
```diff
@@ -217,15 +217,11 @@ var _ = Describe("Reconciling a HostingDeployment that exists", func() {
 	})
 
 	Context("Endpoint does not exist", func() {
```
mbaijal, line 219: Here we mean does not exist in k8s? update message to be specific

RedbackThomson: Agree, this wording was confusing for me too. I'll fix these

surajkota:
reviewed except some unit tests
lgtm

surajkota (approved):
looked at unit tests as well related to deterministic name.

As discussed, we need to address endpointName update/change behaviour which can be separate PR

controllers/hosting/hostingdeployment_controller_test.go
mbaijal: ok for now, but IMO we should review the unit tests for each operator since we understand the operators better. I believe we can clean them up a little bit.

mbaijal (approved):

PR #142, opened 2020-09-16
Fix HAP integration test

akartsky:
Error Message : `[FAILED] HostingDeployment test-namespace:xgboost-hosting job does not exist`

Namespace scoped test was failing after running cluster scoped tests due to replacement of placeholder variables with sed.

### DIFF
--- a/tests/codebuild/create_tests.sh
+++ b/tests/codebuild/create_tests.sh
@@ -139,12 +139,12 @@ function run_hap_test()
   local hostingdeployment_type="hostingdeployment"
 
   # Create the second Endpoint
-  sed -i "s/$hosting_deployment_1/$hosting_deployment_2/g" testfiles/xgboost-hosting-deployment.yaml
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_2
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
 
   # Create the third Endpoint used for the custom metric also here in order to parallize
   # TODO: This test can be written much better by modularizing
-  sed -i "s/$hosting_deployment_2/$hosting_deployment_3/g" testfiles/xgboost-hosting-deployment.yaml
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_3
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
 
   # Endpoints must be created before autoscaling can be applied, one is already created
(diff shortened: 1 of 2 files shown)


tests/codebuild/create_tests.sh
```diff
@@ -139,12 +139,12 @@ function run_hap_test()
...
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_2
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
 
   # Create the third Endpoint used for the custom metric also here in order to parallize
   # TODO: This test can be written much better by modularizing
-  sed -i "s/$hosting_deployment_2/$hosting_deployment_3/g" testfiles/xgboost-hosting-deployment.yaml
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_3
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
 
   # Endpoints must be created before autoscaling can be applied, one is already created
```
RedbackThomson, line 150: Can we replace the `sed` commands for the endpoint metrics, below, as well.

tests/codebuild/create_tests.sh
```diff
@@ -139,12 +139,12 @@ function run_hap_test()
...
 
   # Create the second Endpoint
-  sed -i "s/$hosting_deployment_1/$hosting_deployment_2/g" testfiles/xgboost-hosting-deployment.yaml
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_2
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
 
   # Create the third Endpoint used for the custom metric also here in order to parallize
   # TODO: This test can be written much better by modularizing
-  sed -i "s/$hosting_deployment_2/$hosting_deployment_3/g" testfiles/xgboost-hosting-deployment.yaml
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_3
```
surajkota, line 147: reset to original value at end once test is complete ?

surajkota (approved):

tests/codebuild/create_tests.sh
```diff
@@ -139,12 +139,12 @@ function run_hap_test()
   local hostingdeployment_type="hostingdeployment"
 
   # Create the second Endpoint
-  sed -i "s/$hosting_deployment_1/$hosting_deployment_2/g" testfiles/xgboost-hosting-deployment.yaml
+  yq w -i testfiles/xgboost-hosting-deployment.yaml "metadata.name" $hosting_deployment_2
   run_test "${crd_namespace}" testfiles/xgboost-hosting-deployment.yaml
```
RedbackThomson, line 143: Instead of modifying the test file inline (`yq w -i`), can we just write it to a new file and then apply that?

mbaijal: yq is cleaner in IMO, why do you prefer copying files ?

mbaijal (approved):
Thanks for the fix @akartsky! Looks as good as it can without complete test script refactoring.

tests/codebuild/create_tests.sh
RedbackThomson: I mean instead of using `yq w -i ...` we can do `yq w ... testfiles/xgboost-hosting-deployment-2.yaml` then `run_test ${crd_namespace} testfiles/xgboost-hosting-deployment-2.yaml`. This way we are sure that every time we modify the file, we are modifying the same input file.

akartsky: sorry i didn’t see this message before merging.

i feel we should work on new yamls by default for all the tests and never touch originals 
even when injecting variables here : [inject_tests.sh](https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/tests/codebuild/inject_tests.sh)

but for now yq is fine though

PR #147, opened 2020-10-20
Retain all variant properties

akartsky:
Add RetainAllVariantProperties and ExpectUpdateRequestToHaveRetainVariantProperty to hostingdeployment


tests/codebuild/common.sh
```diff
@@ -52,6 +52,31 @@ function wait_for_crd_status()
...
+  local enpoint_name="$3"
+  local endpoint_region="$4"
+  local timeout="$5"
+  local desired_status="$6"
+
+  echo "Waiting for endpoint $enpoint_name to be in status $desired_status"
+  wait_for_crd_status "${crd_namespace}" HostingDeployment "${crd_instance}" "${timeout}" "${desired_status}"
+
+  timeout "${timeout}" bash -c \
+    'until [ "$(aws sagemaker describe-endpoint --endpoint-name "$0" --region "$1" --query EndpointStatus --output text)" == "$2" ]; do \
```
surajkota, line 69: is this the right apporach? We should be waiting for the k8s resource itself?

What is the root cause of this problem? is the behaviour same if one uses AWS SDK directly instead of operator?

akartsky: there are multiple things which need to be in sync for this test 

this eliminates the possibility of having out of sync status from crd and cli (just for good measure)

surajkota (approved):

mbaijal (approved):
Assuming changes are the same as the private CRI reviewed/approved earlier.

PR #161, opened 2020-12-14
Fix "Containers" parameter in Model operator

akartsky:
Fixes following bugs 
1] Reconciliation failure of Model when `Containers` is specified
2] Model creation failure when `PrimaryContainer` is not specified 

Tested it manually 
Cannot modify existing integ test to include this as `Containers` and `PrimaryContainer` are mutually exclusive.
Need to write new integ test.

### DIFF
--- a/controllers/model/model_controller.go
+++ b/controllers/model/model_controller.go
@@ -33,6 +33,7 @@ import (
 
 	modelv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/model"
 	. "github.com/aws/amazon-sagemaker-operator-for-k8s/controllers"
+	"github.com/aws/amazon-sagemaker-operator-for-k8s/controllers/controllertest"
 	"github.com/aws/amazon-sagemaker-operator-for-k8s/controllers/sdkutil"
 	"github.com/aws/amazon-sagemaker-operator-for-k8s/controllers/sdkutil/clientwrapper"
 )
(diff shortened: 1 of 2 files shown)


surajkota (approved):

controllers/model/model_controller.go
```diff
@@ -33,6 +33,7 @@ import (
 
 	modelv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/model"
 	. "github.com/aws/amazon-sagemaker-operator-for-k8s/controllers"
+	"github.com/aws/amazon-sagemaker-operator-for-k8s/controllers/controllertest"
```
RedbackThomson, line 36: Weird that we are using a test package in a non-test location? Can we move the `ToStringPtr` method to a common package? Maybe `/controllers`?

controllers/model/model_controller.go
```diff
@@ -210,12 +211,14 @@ func (r *ModelReconciler) initializeContext(ctx *reconcileRequestContext) error
 	ctx.SageMakerClient = clientwrapper.NewSageMakerClientWrapper(r.createSageMakerClient(awsConfig))
 	ctx.Log.Info("Loaded AWS config")
 
-	if ctx.Model.Spec.PrimaryContainer.Mode == nil {
-		ctx.Model.Spec.PrimaryContainer.Mode = aws.String(DefaultContainerDefinitionMode)
+	if ctx.Model.Spec.PrimaryContainer != nil {
+		if ctx.Model.Spec.PrimaryContainer.Mode == nil {
```
jkuruba, line 215: Can we combine this to 	if ctx.Model.Spec.PrimaryContainer != nil &&
		 ctx.Model.Spec.PrimaryContainer.Mode == nil 

?

controllers/model/model_controller.go
akartsky: ya that is weird   
it is being used in a lot of places  
i can add this task to our backlog

jkuruba (approved):

PR #164, opened 2021-01-25
Allow multiple container definitions without primary

RedbackThomson:
Allows the user to specify multiple container definitions, without requiring the inclusion of a primary container. SageMaker expects either a single primary container, xor a list of container definitions.

Also, [`gabs`](https://github.com/Jeffail/gabs) was not properly referencing the environment variables for multi-container models. So I have updated the spec converter for this case, and have tested to ensure it works properly.

Fixes #

### All Submissions:


controllers/hosting/model_reconciler.go
```diff
@@ -392,12 +392,17 @@ func (r *modelReconciler) getContainerDefinitions(model *commonv1.Model, contain
...
+		return nil, nil
+	}
+
 	// Get primary container name.
 	// If none is specified and the list of required containers has only one element, use that container as the primary.
 	var primaryContainerName string
 	if model.PrimaryContainer != nil {
 		primaryContainerName = *model.PrimaryContainer
-	} else if len(model.Containers) == 1 {
+	} else if len(model.Containers) > 0 {
```
akartsky, line 405: we should keep this ==1  
much more clearer what is happening

controllers/hosting/model_reconciler.go
```diff
@@ -392,12 +392,17 @@ func (r *modelReconciler) getContainerDefinitions(model *commonv1.Model, contain
...
 		primaryContainerName = *model.Containers[0].ContainerHostname
 	} else {
```
akartsky, line 407: why do we need this "else" ?

RedbackThomson: Although we have a client-side validation for if `len(containers) > 0`, that condition just double checks that this is true. It was one of the conditions of the unit tests.

akartsky: no, i'm talking abt the last else 

```
else {
 return nil, fmt.Errorf("Unable to determine primary ... Either specify explicitly or provide only one container")
}
```

RedbackThomson: Yes that's is also what I'm referring to. In the case that for whatever reason we have no containers, it is a fallback - otherwise we will be doing a null pointer dereference.

controllers/hosting/model_reconciler.go
```diff
@@ -392,6 +392,11 @@ func (r *modelReconciler) getContainerDefinitions(model *commonv1.Model, contain
 // This also returns an error if the primary container definition cannot be found.
 func (r *modelReconciler) getPrimaryContainerDefinition(model *commonv1.Model, containers map[string]*commonv1.ContainerDefinition) (*commonv1.ContainerDefinition, error) {
 
+	// For multi-container, don't expect a primary container
+	if model.PrimaryContainer == nil && len(model.Containers) > 1 {
```
surajkota, line 396: Can we ignore this field if there are more than one containers rather than expecting `model.PrimaryContainer` to be nil? 

because there is no guarantee. from the old code, it looks like one needs to specify one container as primary when using multi-container

controllers/hosting/model_reconciler.go
```diff
@@ -365,16 +365,16 @@ func (r *modelReconciler) getAndValidateContainerMap(model *commonv1.Model) (map
...
 // An error is returned if a required container is not defined.
-func (r *modelReconciler) getContainerDefinitions(model *commonv1.Model, containers map[string]*commonv1.ContainerDefinition, primaryContainerHostname string) ([]*commonv1.ContainerDefinition, error) {
+func (r *modelReconciler) getContainerDefinitions(model *commonv1.Model, containers map[string]*commonv1.ContainerDefinition, primaryContainer *commonv1.ContainerDefinition) ([]*commonv1.ContainerDefinition, error) {
 	var modelContainers []*commonv1.ContainerDefinition
 	for _, container := range model.Containers {
-
-		containerName := *(container.ContainerHostname)
 		// Do not store primary container in containers slice.
-		if containerName == primaryContainerHostname {
+		if container == primaryContainer {
```
surajkota, line 372: can you provide more detail on how this is solving the problem?
to me, it looks like we are ignoring one of the containers from the list

unless I am missing something, it looks like this would be better solved by putting a check for `len(containers)` where these methods are called

surajkota (changes requested):

controllers/hosting/model_reconciler.go
RedbackThomson: Hmm yeah actually that's a good idea. I might just gate it from the outer method rather than modifying these helper methods

controllers/hosting/model_reconciler.go
surajkota: oh I see, primaryContainer will be nil in multi container case. Why do we need this check at all

controllers/hosting/model_reconciler.go
akartsky: in that case can u edit the error message to something like "no containers specified"

controllers/hosting/model_reconciler_test.go
```diff
@@ -386,6 +386,51 @@ var _ = Describe("ModelReconciler.Reconcile", func() {
 			Expect(model.Spec.Tags[0].Value).ToNot(BeNil())
 			Expect(*model.Spec.Tags[0].Value).To(Equal(tagValue))
 		})
+
+		Context("The model has multiple containers", func() {
```
surajkota, line 390: Nice

controllers/model/model_controller.go
```diff
@@ -210,7 +210,7 @@ func (r *ModelReconciler) initializeContext(ctx *reconcileRequestContext) error
 	ctx.SageMakerClient = clientwrapper.NewSageMakerClientWrapper(r.createSageMakerClient(awsConfig))
 	ctx.Log.Info("Loaded AWS config")
 
-	if ctx.Model.Spec.PrimaryContainer.Mode == nil {
+	if ctx.Model.Spec.PrimaryContainer != nil && ctx.Model.Spec.PrimaryContainer.Mode == nil {
```
surajkota, line 213: these changes have already been merged in a previous [PR](https://github.com/aws/amazon-sagemaker-operator-for-k8s/commit/c78903ced42b6a251e4f9bb3347846ce72bbbf71). Any idea why are these showing up?

same for controllers/sdkutil/spec_sdk_converters.go

tests/codebuild/testfiles/xgboost-hosting-deployment-multi-container.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  region: us-west-2
+  productionVariants:
+    - variantName: AllTraffic
+      modelName: xgboost-model
+      initialInstanceCount: 1
+      instanceType: ml.r5.large
+      initialVariantWeight: 1
+  models:
+    - name: xgboost-model
```
surajkota, line 14: we already have 2 hosting deployment tests with and without a name, can we add a container to either of them unless it breaks their assertions?

asking because hosting generates a lot of spam to logs

go.mod
```diff
@@ -25,11 +25,7 @@ require (
 	google.golang.org/protobuf v1.25.0 // indirect
 	k8s.io/api v0.18.8
 	k8s.io/apimachinery v0.18.8
-	k8s.io/cli-runtime v0.18.6 // indirect
 	k8s.io/client-go v0.18.8
-	k8s.io/gengo v0.0.0-20200413195148-3a45101e95ac // indirect
 	k8s.io/klog/v2 v2.2.0 // indirect
 	sigs.k8s.io/controller-runtime v0.6.2
-	sigs.k8s.io/controller-tools v0.3.0 // indirect
```
surajkota, line 33: strange controller-tools is a dependency for generating installers

does this need to be reverted?

controllers/hosting/model_reconciler.go
```diff
@@ -285,13 +285,27 @@ func (r *modelReconciler) extractDesiredModelsFromHostingDeployment(deployment *
...
+			primaryContainer = nil
+			modelContainers = model.Containers
+		} else {
+			// Determine which container is marked as primary
+			if primaryContainer, err = r.getPrimaryContainerDefinition(model, containers); err != nil {
+				return nil, nil, err
+			}
+
+			// Subtract the primary container from the list
+			if modelContainers, err = r.getContainerDefinitions(model, containers, *primaryContainer.ContainerHostname); err != nil {
```
surajkota, line 306: do we need this step if there is only one container?

should modelContainers be set to nil directly?

controllers/model/model_controller.go
RedbackThomson: Oh odd. I mustn't be rebased maybe

go.mod
RedbackThomson: It's a dependency as a tool we need, it's not a dependency for the code itself (hence it was only indirect). I could add it as a direct dependency, maybe. Still not certain how these modules are installed and operate.

akartsky (commented):
Looks good 
once Suraj's comments are resolved let me know

surajkota (approved):

PR #167, opened 2021-01-29
fix: hap test uses named resource causes race condition for canaries

surajkota:
HAP test use a named endpoint causing race condition if more than one test suite is running in parallel. Hence, moving it to integration tests
To enable back in canaries test will need to be modified

### All Submissions:

### DIFF
--- a/tests/codebuild/update_tests.sh
+++ b/tests/codebuild/update_tests.sh
@@ -9,12 +9,6 @@ source inject_tests.sh
 function run_update_canary_tests
 {
   local crd_namespace="$1"
-
-  echo "Injecting variables into tests"
-  inject_all_variables
-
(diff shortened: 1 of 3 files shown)


akartsky (approved):

tests/codebuild/update_tests.sh
```diff
@@ -9,12 +9,6 @@ source inject_tests.sh
 function run_update_canary_tests
 {
   local crd_namespace="$1"
-
-  echo "Injecting variables into tests"
-  inject_all_variables
```
RedbackThomson, line 14: Inject should stay here. Eventually when we add update tests for canaries, they'll run here first. The integration tests always run the `run_update_canary_tests` function too.

PR #169, opened 2021-02-08
Revert PR #167 and randomize endpoint name in HAP test

mbaijal:
1. This reverts commit 24cd54bdf041d6e347f7d8877460ed02805b0432 (PR #167 )
2. adds a random string to the endpoint name to fix the race condition (both create and update test)

### [WIP] Testing
1. Reproduced the error by reverting PR #167 and triggering two integration test runs to see that one failed. 
2. Used the same steps to then check that the fix actually works and the race condition is not reproduced. 

### All Submissions:


tests/codebuild/update_tests.sh
```diff
@@ -57,8 +58,9 @@ function verify_update_integration_tests
 #    $3: Filename of the hap test to update
 function update_hap_test()
 {
+  local random_string=$RANDOM
```
RedbackThomson, line 61: Where does this come from?

tests/codebuild/update_tests.sh
```diff
@@ -9,6 +9,12 @@ source inject_tests.sh
 function run_update_canary_tests
 {
   local crd_namespace="$1"
+
+  echo "Injecting variables into tests"
```
surajkota, line 13: I think we should revert changes to this file when we revert #155 so that we have this test run through integ before canary.

just being cautious for canaries here

surajkota (changes requested):

tests/codebuild/create_tests.sh
```diff
@@ -200,8 +200,9 @@ function verify_retain_varient_properties(){
 #    $4: Filename of the custom metric hap test to run
 function run_hap_test()
 {
+  local random_string=$RANDOM
   local target_namespace="$1"
-  local hosting_deployment_1="$2"
+  local hosting_deployment_1="${2}-${random_string}"
```
surajkota, line 205: line 232 in this method resets the endpoint name in the yaml file after the resources are applied. we need to use a different variable for that now?

also where is $RANDOM generated

tests/codebuild/update_tests.sh
mbaijal: I found this in in several unofficial docs such as - 
1. https://www.geeksforgeeks.org/random-shell-variable-in-linux-with-examples/
2. https://stackoverflow.com/questions/8988824/generating-random-number-between-1-and-10-in-bash-shell-script

Other alternatives included using tools such as `shuf` or `tr` or using the official random number generator `/dev/random` (or `/dev/urandom`). But I tested the `RANDOM` shell variable locally as well as in the pipeline and it worked fine. We don't need it to be perfectly sampled etc, so it should be ok. 
 
Do you see any issues with using this ?

RedbackThomson: Wow I've never heard of this before. That's super cool. There is still a chance that we have collisions between runs, though? I'm not sure if there is enough entropy in this alone?

tests/codebuild/create_tests.sh
mbaijal: Since we are using `yq` everywhere now and not `sed` the actual value in these fields should not matter. However, I will make sure to parse the logs of a test run (including the namespace tests) to make sure everything actually ran as expected.

tests/codebuild/update_tests.sh
mbaijal: Ok, let me revert PR 155 as well and fix the underlying issue.

mbaijal: Done

surajkota:
Closing since, it is not being actively worked on

Issue #165, opened 2021-01-26
Upgrade aws-sdk to v1


EngHabu:
**What would you like to be added**:
Upgrade dependency to the stable aws-sdk v1.. it's a breaking change so a lot will need to be updated.

**Why is this needed**:
Unable to consume any of new SDK types because we have a dependency on the operator.

mbaijal:
Hello Haytham, 
Thank you for your request. We do not have a decision on this currently and we'll get back to you. 

Meghna Baijal

surajkota:
Please update to v1.2.0 which has the changes requested.

PR #171, opened 2021-02-15
Migrate to a stable version of AWS GoSDK V1

akartsky:
Migrate from pre-release version of AWS GoSDK V2 to a stable version of AWS GoSDK V1 
`from [aws-sdk-go-v2 v0.24.0]  to [aws-sdk-go v1.37.3]` 

Move to a stable version of GoSDK 

- ~~3 Unit tests are skipped in BatchTransform related to custom endpoint. Working on resolving the issue.~~ FIXED
- smlog tests are skipped but the package isn't deleted yet 

Yes
Working with Doc team 

-------------------------------------------------------------

## Testing 

All integration tests are passing 

All Unit tests are passing ~~except the 3 pending tests~~
```
:k8s_operators kalamadi$ go test -v -short ./controllers/... | grep Passed
SUCCESS! -- 16 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 16 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 20 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 101 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 21 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 20 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 44 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 33 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
SUCCESS! -- 85 Passed | 0 Failed | 0 Pending | 0 Skipped
```


akartsky:
https://github.com/aws/amazon-sagemaker-operator-for-k8s/issues/165

api/v1/common/sagemaker_api.go
```diff
@@ -408,7 +408,7 @@ type ObjectiveStatus string
 // Hosting related types
 
 type ProductionVariant struct {
-	AcceleratorType ProductionVariantAcceleratorType `json:"acceleratorType,omitempty"`
+	AcceleratorType *string `json:"acceleratorType,omitempty"`
```
RedbackThomson, line 411: Why not make them `*ProductionVariantAcceleratorType` ? These should be identical

controllers/aws_config_loader.go
```diff
@@ -42,15 +41,24 @@ func NewAwsConfigLoaderForEnv(env venv.Env) AwsConfigLoader {
 	}
 }
 
+// CreateNewAWSSessionFromConfig returns an AWS session using AWS Config
+func CreateNewAWSSessionFromConfig(cfg aws.Config) *awssession.Session {
+	sess, _ := awssession.NewSessionWithOptions(
+		awssession.Options{
+			SharedConfigState: awssession.SharedConfigEnable,
```
RedbackThomson, line 48: Could you link, in a comment above, to the issue which describes why we need this flag?

controllers/aws_config_loader.go
```diff
@@ -42,15 +41,24 @@ func NewAwsConfigLoaderForEnv(env venv.Env) AwsConfigLoader {
 	}
 }
 
+// CreateNewAWSSessionFromConfig returns an AWS session using AWS Config
+func CreateNewAWSSessionFromConfig(cfg aws.Config) *awssession.Session {
```
RedbackThomson, line 45: nit: Line 57 uses `Aws`. I would personally prefer all references to AWS be `AWS` and not `Aws`. But if that's too much work, lets stick to the existing use.

controllers/aws_config_loader.go
```diff
@@ -42,15 +41,24 @@ func NewAwsConfigLoaderForEnv(env venv.Env) AwsConfigLoader {
...
 // LoadAwsConfigWithOverrides loads default AWS config and apply overrides, like setting the region and using a custom SageMaker endpoint.
 // If specified, jobSpecificEndpointOverride always overrides the endpoint. Otherwise, the environment
 // variable specified by DefaultSageMakerEndpointEnvKey overrides the endpoint if it is set.
 func (l AwsConfigLoader) LoadAwsConfigWithOverrides(regionOverride string, jobSpecificEndpointOverride *string) (aws.Config, error) {
 	var config aws.Config
-	var err error
 
-	if config, err = external.LoadDefaultAWSConfig(external.WithRegion(regionOverride)); err != nil {
-		return aws.Config{}, err
+	if regionOverride != "" {
```
RedbackThomson, line 60: Comparison to empty string is an antipattern for checking if something should exist. Perhaps changing `regionOverride` to `*string` and then checking `regionOverride != nil` would be cleaner.

controllers/aws_config_loader.go
```diff
@@ -64,17 +72,17 @@ func (l AwsConfigLoader) LoadAwsConfigWithOverrides(regionOverride string, jobSp
 
 	// If a custom endpoint is requested, install custom resolver for SageMaker into config.
 	if customEndpoint != "" {
-		customSageMakerResolver := func(service, region string) (aws.Endpoint, error) {
-			if service == sagemaker.EndpointsID {
-				return aws.Endpoint{
+		customSageMakerResolver := func(service, region string, optFns ...func(*awsendpoints.Options)) (awsendpoints.ResolvedEndpoint, error) {
+			if service == "sagemaker" {
+				return awsendpoints.ResolvedEndpoint{
```
RedbackThomson, line 77: Reminder that we need to run PrivateLink tests

controllers/batchtransformjob/batchtransformjob_controller.go
```diff
@@ -65,8 +66,8 @@ func NewBatchTransformJobReconciler(client client.Client, log logr.Logger, pollI
 		Client:       client,
 		Log:          log,
 		PollInterval: pollInterval,
-		createSageMakerClient: func(cfg aws.Config) sagemakeriface.ClientAPI {
-			return sagemaker.New(cfg)
+		createSageMakerClient: func(cfg aws.Config) sagemakeriface.SageMakerAPI {
+			return sagemaker.New(CreateNewAWSSessionFromConfig(cfg))
```
RedbackThomson, line 70: ++ Clean

controllers/batchtransformjob/batchtransformjob_controller.go
```diff
@@ -388,7 +389,7 @@ func (r *BatchTransformJobReconciler) getSageMakerDescription(ctx reconcileReque
 		log.Info("Failed to parse the describe error output from sagemaker")
 		return nil, awsErr
 	}
-	return describeResponse.DescribeTransformJobOutput, nil
+	return describeTransformJobOutput, nil
```
RedbackThomson, line 392: Does this return the response of the describe? It seems to be returning the response of the `DescribeTransformJobRequest`? That means we don't have to wait for `describeRequest.Send()`?

controllers/controllertest/mock_autoscaling_client.go
```diff
@@ -287,23 +288,16 @@ func (m mockApplicationAutoscalingClient) DescribeScalableTargetsRequest(input *
...
-			r.Data = nextdescribeScalableTargetsResponse.targetData
-		})
+		mockRequest.Error = nextdescribeScalableTargetsResponse.err
+		return mockRequest, nil
 	}
 
-	return applicationautoscaling.DescribeScalableTargetsRequest{
-		Request: mockRequest,
-	}
+	return mockRequest, nextdescribeScalableTargetsResponse.targetData
```
RedbackThomson, line 295: Much cleaner ++

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
```diff
@@ -80,14 +80,19 @@ func (s hpoTrainingJobSpawner) SpawnMissingTrainingJobs(ctx context.Context, hpo
...
-
-	paginator := s.SageMakerClient.ListTrainingJobsForHyperParameterTuningJob(ctx, hpoJobName)
+	var nextToken *string = nil
 
 	// WaitGroup allowing us to do checks in parallel.
 	var wg sync.WaitGroup
 
-	for paginator.Next(ctx) {
-		list := paginator.CurrentPage()
+	for {
```
RedbackThomson, line 88: Where is the condition for this `for`? Is this the equivalent of a `while True`?

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
```diff
@@ -80,14 +80,19 @@ func (s hpoTrainingJobSpawner) SpawnMissingTrainingJobs(ctx context.Context, hpo
...
+		hpoTrainingJobOutput, err := s.SageMakerClient.ListTrainingJobsForHyperParameterTuningJob(ctx, hpoJobName, nextToken)
+		if err != nil {
+			s.Log.Info("Error while getting training jobs", "err", err)
```
RedbackThomson, line 91: nit: Should be `s.Log.Error`.

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
```diff
@@ -112,10 +117,12 @@ func (s hpoTrainingJobSpawner) SpawnMissingTrainingJobs(ctx context.Context, hpo
...
 		}
-	}
 
-	if err := paginator.Err(); err != nil {
-		s.Log.Info("Error while getting training jobs", "err", err)
+		if hpoTrainingJobOutput.NextToken == nil {
+			break
+		} else {
+			nextToken = hpoTrainingJobOutput.NextToken
+		}
```
RedbackThomson, line 125: random nit: Don't need `else`. `break` will already exit scope.

RedbackThomson (changes requested):
I am actually super excited about these changes! It looks like a lot, but when you ignore the pointer reference/dereference, it's actually not too bad. 
Great work!

api/v1/common/sagemaker_api.go
akartsky: `string` seems more straight forward

controllers/aws_config_loader.go
akartsky: https://github.com/aws/aws-sdk-go/issues/3101

https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

controllers/aws_config_loader.go
akartsky: The default value for `string` variable is "" so it seems a reasonable check to me.

Any string reason why we should not do this ?

controllers/batchtransformjob/batchtransformjob_controller.go
akartsky: yes it does return the response of DescribeTransformJobRequest.

it will only be filled after doing the `describeRequest.Send()` that's how the API works

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
akartsky: yep

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
akartsky: lol

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
akartsky: Nice catch

akartsky:
> I am actually super excited about these changes! It looks like a lot, but when you ignore the pointer reference/dereference, it's actually not too bad.
> Great work!

Thanks 😄

Oh the pointer reference/dereference is deceptively bad 
coz of all the runtime issues which take a lot of time to debug 
(279 unit test cases were failing)

just to give a few examples 
```
# this `if` condition is not satisfied if a and b go from string to *string
if a==b {}
# and this could cause problems in totally different part of code 
# and only way we debug this is to go through the whole flow of the unit test case and add print statements everywhere 

# append strings to array 
# now if a goes from string to *string, same pointer is appended and the whole array ends up having the same value 
for a in b.send()
    egArray = append(egArray, a)
``` 

There are many of these kinda errors which are hard to catch and take lot of time.

controllers/aws_config_loader.go
akartsky: btw we do have integ test for custom endpoint which are passing

controllers/aws_config_loader.go
RedbackThomson: I meant a comment in the code

controllers/aws_config_loader.go
RedbackThomson: We pass `jobSpecificEndpointOverride` as a pointer but `regionOverride` as a value. Typically if we mean "no value" then, in Go at least, we want to pass a `nil` pointer - rather than a "". This is the equivalent of passing `None` instead of `""` in Python.

controllers/batchtransformjob/batchtransformjob_controller.go
RedbackThomson: Super weird. Okay

api/v1/common/sagemaker_api.go
RedbackThomson: Yes but we lose specificity when we make this change. We have lots of other types in this file that alias string, even `ProductionVariantInstanceType` is still an alias. This change isn't relevant to Go SDK v1 migration, so I would like to keep it the same.

controllers/aws_config_loader.go
akartsky: still not sold on empty string comparison 
but for consistency changing `regionOverride` to pointer and aws.Config expects it to be pointer anyway

controllers/aws_config_loader.go
akartsky: ya i agree, `AWS` is better 
changed all the functions to use that acronym

controllers/aws_config_loader.go
akartsky: done

api/v1/common/sagemaker_api.go
akartsky: * yes we have other alias which are `string` but none of them are passed as pointers `*alias`
  * In this file we have 18 string type variables and all are passed as string but never as string pointer 
* All `*string` variables are passed as is and they do not have aliases 
  * In this file we have 79 `*string` variables all of which are passed as `*string` and they don't have alias
* This change is relevant to Go SDK v1 migration because `ProductionVariantAcceleratorType` went from being a `string` to `*string`. This datatype is used in the controllers.

So this change of removing `ProductionVariantAcceleratorType` and using `*string` is consistent with what we currently have in this file

RedbackThomson: Absolutely fair points. Thank you for explaining this to me

controllers/hyperparametertuningjob/hpo_trainingjob_spawner.go
```diff
@@ -248,19 +261,13 @@ func (s hpoTrainingJobSpawner) deleteSpawnedTrainingJobsConcurrently(ctx context
...
-	// blocked before it consumes from the channel.
-	wg.Add(1)
-	go func(paginatorError error) {
-		defer wg.Done()
-		if paginatorError != nil {
-			s.Log.Info("Error while getting training jobs", "err", paginatorError)
-			errorsChannel <- paginatorError
+		if hpoTrainingJobOutput.NextToken == nil {
+			break
+		} else {
```
surajkota, line 267: nit: eliminate else ?

controllers/controllertest/mock_sagemaker_client.go
```diff
@@ -682,23 +675,16 @@ func (m mockSageMakerClient) DescribeTrainingJobRequest(input *sagemaker.Describ
...
 
 	if nextDescribeTrainingJobResponse.err != nil {
-		mockRequest.Handlers.Send.PushBack(func(r *aws.Request) {
-			r.Error = nextDescribeTrainingJobResponse.err
-		})
-	} else {
-		mockRequest.Handlers.Send.PushBack(func(r *aws.Request) {
-			r.Data = nextDescribeTrainingJobResponse.data
-		})
+		mockRequest.Error = nextDescribeTrainingJobResponse.err
```
surajkota, line 678: can you explain this change a bit more?
I thought this is needed so aws client in controller could consume the response. Idea is we push mock responses and when controller pops them per API call instead of making actual API calls

surajkota (commented):
Great work Kartik!!
Looks scary 47 pages but is pretty clean and self explainatory

controllers/batchtransformjob/batchtransformjob_controller_test.go
```diff
@@ -741,12 +745,12 @@ var _ = Describe("Reconciling when a custom SageMaker endpoint is requested", fu
...
 
 		Expect(actualEndpointResolver).ToNot(BeNil())
-		actualEndpoint, err := (*actualEndpointResolver).ResolveEndpoint(sagemaker.EndpointsID, uuid.New().String())
+		actualEndpoint, err := (*actualEndpointResolver).EndpointFor(sagemaker.EndpointsID, uuid.New().String())
 		Expect(err).ToNot(HaveOccurred())
 		Expect(actualEndpoint.URL).To(Equal(expectedEndpoint))
 	})
 
-	It("should use the job-specific SageMakerEndpoint over the environment variable", func() {
+	PIt("should use the job-specific SageMakerEndpoint over the environment variable", func() {
```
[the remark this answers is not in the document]
akartsky: Pending test ... these will be skipped 

https://github.com/onsi/ginkgo/blob/v1.15.0/ginkgo_dsl.go#L394

controllers/trainingjob/trainingjob_controller.go
```diff
@@ -224,15 +225,15 @@ func (r *Reconciler) reconcileTrainingJob(ctx reconcileRequestContext) error {
...
 	}
 
-	primaryStatus := string(ctx.TrainingJobDescription.TrainingJobStatus)
-	secondaryStatus := string(ctx.TrainingJobDescription.SecondaryStatus)
+	primaryStatus := *ctx.TrainingJobDescription.TrainingJobStatus
+	secondaryStatus := *ctx.TrainingJobDescription.SecondaryStatus
 	additional := controllers.GetOrDefault(ctx.TrainingJobDescription.FailureReason, "")
 
-	if ctx.TrainingJobDescription.TrainingJobStatus == sagemaker.TrainingJobStatusStopping {
+	if *ctx.TrainingJobDescription.TrainingJobStatus == string(sagemaker.TrainingJobStatusStopping) {
```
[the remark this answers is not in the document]
akartsky: `sagemaker.TrainingJobStatusStopping` is of type *string

can't compare string pointers so converting both sides to string

controllers/trainingjob/trainingjob_controller_test.go
```diff
@@ -281,8 +281,8 @@ var _ = Describe("Reconciling a TrainingJob that exists", func() {
 
 	Context("TrainingJob exists", func() {
 
-		var expectedStatus sagemaker.TrainingJobStatus
-		var expectedSecondaryStatus sagemaker.SecondaryStatus
+		var expectedStatus string
```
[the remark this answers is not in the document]
akartsky: string is the type 

they used to have specific types but now they don't

controllers/controllertest/mock_sagemaker_client.go
akartsky: main purpose of this block is to return "error" or "data" according to the test case 

Error is return by simply setting the r.Error 
which will be checked during request.Send() by sagemaker api
https://github.com/aws/aws-sdk-go/blob/master/aws/request/request.go#L518

Data is return directly on line 682

tests/codebuild/run_all_sample_test.sh
```diff
@@ -15,6 +15,6 @@ verify_integration_tests ${crd_namespace}
 # run_update_integration_tests ${crd_namespace}
 # verify_update_integration_tests ${crd_namespace}
 verify_feature_integration_tests ${crd_namespace}
-run_smlogs_integration_tests ${crd_namespace}
+# run_smlogs_integration_tests ${crd_namespace}
```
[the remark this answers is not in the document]
akartsky: need to make the changes to the smlog so that the package compiles 

tests are removed coz we don't actually need smlog and i didn't resolve runtime errors

controllers/batchtransformjob/batchtransformjob_controller.go
```diff
@@ -112,7 +113,7 @@ func (r *BatchTransformJobReconciler) reconcileJob(ctx reconcileRequestContext)
 		return RequeueImmediately()
 	}
 
-	if awsConfig, err := r.awsConfigLoader.LoadAwsConfigWithOverrides(*ctx.Job.Spec.Region, ctx.Job.Spec.SageMakerEndpoint); err != nil {
+	if awsConfig, err := r.awsConfigLoader.LoadAwsConfigWithOverrides(ctx.Job.Spec.Region, ctx.Job.Spec.SageMakerEndpoint); err != nil {
```
RedbackThomson, line 116: nit: `LoadAwsConfigWithOverrides` -> `LoadAWSConfigWithOverrides`

controllers/endpointconfig/endpointconfig_controller.go
```diff
@@ -67,10 +67,10 @@ func NewEndpointConfigReconciler(client client.Client, log logr.Logger, pollInte
 		Client:       client,
 		Log:          log,
 		PollInterval: pollInterval,
-		createSageMakerClient: func(awsConfig aws.Config) sagemakeriface.ClientAPI {
-			return sagemaker.New(awsConfig)
+		createSageMakerClient: func(cfg aws.Config) sagemakeriface.SageMakerAPI {
```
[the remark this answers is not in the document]
RedbackThomson: This is repeated for the sake of mocking in unit tests. Is that what you mean?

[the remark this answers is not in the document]
akartsky: abstracting it would be nice 
but some of them use client wrappers 

https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/controllers/trainingjob/trainingjob_controller.go#L65

don't want to get into that as this seems unrelated code refactor

controllers/batchtransformjob/batchtransformjob_controller_test.go
```diff
@@ -642,6 +644,8 @@ var _ = Describe("Reconciling a job without finalizer that is being deleted", fu
 	})
 })
 
+// Currently these 3 tests are skipped
```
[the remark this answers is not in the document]
akartsky: ya it was until now 😅
Fixed

jkuruba (commented):
Looks to be a clean code mapping for versions. Did we test private links as requested in another comment?

tests/codebuild/run_all_sample_test.sh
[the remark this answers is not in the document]
surajkota: rather than making changes to the package, deleting the tests, is it possible to comment out the compilation for sake of this PR?

basically, I am looking to commenting out everything (build, test, release) from the pipeline related to smlogs in one PR. right now its partial?

akartsky:
> Looks to be a clean code mapping for versions. Did we test private links as requested in another comment?

thanks 😄

private link test is not done yet

tests/codebuild/run_all_sample_test.sh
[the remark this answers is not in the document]
akartsky: fixed smlog

surajkota (approved):

akartsky:
FYI 
private link testing done before merging

PR #173, opened 2021-02-23
Login to dockerhub

akartsky:

Login to dockerhub to avoid hitting the limit

### DIFF
--- a/codebuild/build_image.yaml
+++ b/codebuild/build_image.yaml
@@ -1,5 +1,8 @@
 version: 0.2      
 phases:
+  pre_build:
+    commands:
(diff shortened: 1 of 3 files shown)


codebuild/build_image.yaml
```diff
@@ -1,5 +1,8 @@
 version: 0.2      
 phases:
+  pre_build:
```
surajkota, line 3: do we also need to add it to unit test, build_canary and deploy_image as well ?

deploy_image also fails sometimes so yes for that. not sure about the rest

surajkota (approved):

PR #192, opened 2021-12-14
go mod tidy

akartsky:

- Fix bug in smlogs go.mod file (added aws go sdk version v1 and removed v2)
- Ran `go mod tidy` to remove unused packages 
- Modify `Makefile` to use `go install` instead of `go get` so that the go.mod and go.sum files don't change (This was causing the unit tests to fail)


smlogs-kubectl-plugin/go.mod
```diff
@@ -4,15 +4,13 @@ go 1.13
 
 require (
 	github.com/aws/amazon-sagemaker-operator-for-k8s v0.0.0
-	github.com/aws/aws-sdk-go-v2 v0.24.0
+	github.com/aws/aws-sdk-go v1.37.3
```
surajkota, line 7: is this going to work out of the box?

akartsky: This was a miss. It should have always been v1.37.3

I believe the reason it was working before is because we build SageMaker operators first which would install v1 
Not sure

surajkota: so do we even need this go mod? can you try removing it?
this should have been a separate go package

akartsky: It is required. The behavior of the "go" changed thats what revealed this problem.
Old version was go1.15.2
Current version is go.1.17.5

https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/tests/images/Dockerfile.integration#L48
The version isn't pinned in the docker container that is used in tests.
This container is not built every time. I had to build it only to upgrade the `eksctl`.

ryansteakley (approved):

PR #200, opened 2022-08-19
Force Delete HPO Training Jobs

mbaijal:
Force deletes trainingjobs thus making sure that test runs complete. 

Tested: 
1. The branch on the integration test pipeline
2. Locally tested on the nonephemeral cluster that all stuck jobs get cleaned up. 

### DIFF
--- a/tests/codebuild/common.sh
+++ b/tests/codebuild/common.sh
@@ -54,6 +54,19 @@ function wait_for_crd_status()
   fi
 }
 
+function force_delete_training_jobs()
+{
+  training_jobs=$(kubectl get trainingjobs -ojson | jq -r '.items | .[] | .metadata.name')
+ 
+for job in $training_jobs
+do
+    echo $job
+    kubectl patch trainingjob $job -p '{"metadata":{"finalizers":null}}' --type=merge
+done
+ 
+kubectl delete trainingjob --all
+}
+
 # Cleans up all resources created during tests.
 # Parameter:
 #    $1: Namespace of CRD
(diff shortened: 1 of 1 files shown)


tests/codebuild/common.sh
```diff
@@ -54,17 +54,22 @@ function wait_for_crd_status()
...
  
 for job in $training_jobs
 do
     echo $job
-    kubectl patch trainingjob $job -p '{"metadata":{"finalizers":null}}' --type=merge
+    kubectl patch -n "$crd_namespace" trainingjob $job -p '{"metadata":{"finalizers":null}}' --type=merge
 done
  
-kubectl delete trainingjob --all
+kubectl delete -n "$crd_namespace" hyperparametertuningjob --all
```
surajkota, line 71: is this needed? I think hpo jobs get deleted fine

tests/codebuild/common.sh
```diff
@@ -83,7 +88,7 @@ function delete_all_resources()
   kubectl delete -n "$crd_namespace" hostingdeployment --all 
   kubectl delete -n "$crd_namespace" model --all 
 
-  force_delete_training_jobs 
+  force_delete_training_jobs "$crd_namespace"
```
surajkota, line 91: this will only work if we add a timeout for the first trainingjob delete line 82

surajkota (changes requested):

tests/codebuild/common.sh
mbaijal: I actually went back and added this just in case. No harm in being thorough if we are force deleting.

tests/codebuild/common.sh
mbaijal: Why ? I can add some sleep in there.

surajkota: because each command here runs serially and would get stuck at line 82 unless there is a exit condition like `kubectl delete --wait=60sec`

surajkota:
closing in favour of #204

